PR #12, opened 2026-02-19
Add support for scala

pityka:
I mostly followed Java's treatment.

### DIFF
--- a/src/read/outline/code.rs
+++ b/src/read/outline/code.rs
@@ -320,7 +320,12 @@ fn extract_doc(node: tree_sitter::Node, lines: &[&str]) -> Option<String> {
 }
 
 /// Format outline entries into the spec'd output format.
-fn format_entries(entries: &[OutlineEntry], _lines: &[&str], max_lines: usize, lang: Lang) -> String {
+fn format_entries(
+    entries: &[OutlineEntry],
+    _lines: &[&str],
+    max_lines: usize,
+    lang: Lang,
+) -> String {
     let mut out = Vec::new();
     let mut import_groups: Vec<&str> = Vec::new();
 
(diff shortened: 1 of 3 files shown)


jahala:
Thank you for that, much appreciated 🙏🏻  And thank you for the update on the PR .. here is what Opus 4.6 has to say about it: 

----

Great update — this addresses all the major concerns from the first review. The Scala-specific tree-sitter queries are correct (`call_expression`, `field_expression`, `infix_expression`), the outline covers all the key constructs, and the tests are a nice addition.

A few small things before merging:

**Bug — `var` displays as `val` in outline:**

The code maps both `val_definition` and `var_definition` to `OutlineKind::Variable`, and the label always renders as `"val"` for Scala. So this:
```scala
var connected = false
```
shows up as `val connected` in the outline. The test on line 564 actually asserts this incorrect behavior. Since `val` (immutable) vs `var` (mutable) is a meaningful distinction in Scala, this should be fixed — either by adding a separate `OutlineKind` variant, or by checking the node kind in `format_entry`.

**Minor nits:**
- Missing trailing comma on `"build.sbt"` (line 86 in search/mod.rs) — inconsistent with the other entries
- `println!` debug output left in tests (callees.rs:476, siblings.rs:249) — should be removed

Everything else looks good. The `target` addition to SKIP_DIRS makes sense for both Scala and Rust projects. Happy to merge once these are addressed.

pityka:
Hello! I made a second round on these. scala vals are assigned to a new outline kind named ImmutableVariable. I believe rust's let variable should go into this as well but I did not change anything on that side. I opted for a new kind rather than reusing Constant because as far as I understood Constant is for compile time constants.

PR #28, opened 2026-03-11
Add PHP tree-sitter support across search and outlines

quanhavn:
## Summary
- add Lang::Php and detect .php/.phtml files as code
- wire tree-sitter-php into outline_language and handle PHP declaration/import/namespace node kinds in outline extraction
- extend structural definition/caller/callee logic with PHP node/query patterns
- add PHP-focused tests for outline generation and callee extraction
- update README and CLAUDE language support docs

## Validation
- cargo test


jahala:
Hey @quanhavn

First of all thank you so much for your contribution! I had Opus 4.6 have a look, and this is the feedback: 

---

Thanks for the contribution! Clean work — follows the existing patterns well.

Reviewed the diff:

* tree-sitter-php 0.24.2 — official tree-sitter org crate, correct. LANGUAGE_PHP (not LANGUAGE) is the right choice for PHP-with-HTML.
* Node kinds (trait_declaration, const_declaration, namespace_definition, namespace_use_declaration, nullsafe_member_call_expression) all match the tree-sitter-php grammar.
* namespace_definition in is_namespace — correct, PHP namespaces should be transparent wrappers like C#.
* first_identifier_text fallback for const_declaration — good, PHP const nodes don’t expose a name field the same way Rust does.
* Callee queries cover plain calls, method calls, nullsafe calls, static calls, and qualified names. Tests verify all four.


A few things that could be follow-ups (not blockers for this PR):
1. Sibling query — $this->method() / $this->field patterns aren’t wired up yet, so sibling hints won’t appear for PHP definitions. Would be something like (member_access_expression object: (variable_name (name) @obj) name: (name) @ref) with a $this filter.
2. Import resolution — is_import_line doesn’t handle PHP use statements, so tilth_deps won’t resolve PHP imports. Same gap as Ruby/Swift currently.

Neither blocks merging — PHP lands at the same support level as Ruby/C/C++.

PR #32, opened 2026-03-15
feat: add Amp (ampcode.com) as supported MCP host

QuangThai:
## Summary
- add Amp (`ampcode.com`) as a supported MCP host for `tilth install`
- write Amp config to `~/.config/amp/settings.json`
- store the server entry under the literal `amp.mcpServers` key, matching Amp's config format
- preserve unrelated config while overwriting any existing `tilth` entry
- add tests covering overwrite, preservation, invalid existing config, and the dotted-key edge case

## Notes
- Also includes a small pre-existing fix in `src/main.rs`: the `tilth install` help text now correctly lists `gemini` and `codex` among the supported hosts. This is unrelated to the Amp config change, but it was worth fixing while touching the install surface.

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -280,11 +280,7 @@ fn home_dir() -> Result<PathBuf, String> {
 
 /// Merge a tilth server entry into a JSON config under the given servers key.
 /// Extracted for testability — used by `write_json_config` and unit tests.
-fn upsert_json_server(
-    config: &mut Value,
-    servers_key: &str,
-    entry: Value,
-) -> Result<(), String> {
+fn upsert_json_server(config: &mut Value, servers_key: &str, entry: Value) -> Result<(), String> {
     config
         .as_object_mut()
         .ok_or("config root is not a JSON object")?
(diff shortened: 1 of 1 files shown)


jahala:
Hey @QuangThai 

Thank you so much for your PR! Sorry about the late reply, I've been travelling for the last week. 

Here's what Opus 4.6 has to say about your PR: 

---


Thanks for the contribution! Config path and "amp.mcpServers" key format both check out against the official Amp docs.

A couple of small notes:

* The help text fix in main.rs (adding gemini, codex) is a good catch — that was a pre-existing bug. Could you mention it in the PR description so it’s not a silent drive-by?

* The #[derive(Debug)] additions on ConfigFormat and HostInfo — are these needed? They don’t appear to be used by the tests or the feature. Happy to keep them if there’s a reason, but otherwise prefer to keep the diff minimal.

Otherwise LGTM — clean code, good tests, especially the dotted-key edge case. Will merge once those are addressed.

QuangThai:
Hi @jahala 

Thanks for catching that. The failure was just a formatting issue in `src/install.rs`.

I have run `cargo fmt`, pushed the fix in `b33e9a0`, and verified that `cargo fmt --check` and `cargo test install` now pass locally.

PR #48, opened 2026-03-31
feat: show compact diff in tilth_edit response

QuangThai:
## Summary

`tilth_edit` now returns a **compact diff** showing exactly what changed at each edit site, solving the problem where agents and users could not see before/after changes without re-reading the file.

### Before

```
42:b7e|  let x = recompute();
43:c2d|  let y = validate(x);
44:a1f|  return y;
```
Only hashlined context around the edit site (post-edit state). No visibility into what was removed or replaced.

### After

```
── diff ──
:42-44
- 42:a3f|  let x = compute();
- 43:f1b|  return x;
+ 42:b7e|  let x = recompute();
+ 43:c2d|  let y = validate(x);
+ 44:a1f|  return y;

42:b7e|  let x = recompute();
43:c2d|  let y = validate(x);
44:a1f|  return y;
```
Diff block with `-`/`+` lines precedes the existing context. Hashline anchors on `+` lines are immediately usable for follow-up edits.

## Design decisions

1. **Inline diff in tool response** — agent sees changes immediately, no extra tool call
2. **Compact tilth-style format** — `-`/`+` with hashline anchors, not full unified diff; token-efficient
3. **Per-edit-site blocks** — each edit in a batch gets its own diff section
4. **Offset-adjusted `+` line numbers** — when earlier edits in a batch insert/delete lines, `+` lines show correct post-edit positions matching the written file
5. **Opt-out via `diff: false`** — agents can suppress diff output to save tokens
6. **Additive, backward compatible** — diff prepends existing context + blast radius output; no breaking changes

### `src/edit.rs`
- `EditResult::Applied` now carries `{ diff, context }` instead of a single string
- `EditDiff` struct with `old_start`/`new_start` (original vs offset-adjusted positions)
- `old_snapshots` captured before edits are applied (for before/after comparison)
- `format_diffs()` renders `── diff ──` header + per-edit `-`/`+` blocks
- 10 unit tests: single-line, multi-line (fewer/more lines), deletion, batch, empty, header format, unicode, offset-adjusted batch deletion, offset-adjusted batch insertion

### `src/mcp.rs`
- `tool_edit()` reads new `diff` boolean param (default `true`), assembles response as: diff → context → blast radius
- `tilth_edit` tool schema includes `diff` parameter
- `EDIT_MODE_EXTRA` instructions updated

### `AGENTS.md`
- One line added documenting diff output in edit responses

## Verification

- `cargo fmt --check` ✓
- `cargo clippy -- -D warnings` ✓
- `cargo test` — 108 tests pass (98 existing + 10 new) + 1 doctest ✓
- No new dependencies

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -47,6 +47,6 @@ Range: {"start": "<line>:<hash>", "end": "<line>:<hash>", "content": "..."}
 Delete: {"start": "<line>:<hash>", "content": ""}
 Hash mismatch → file changed, re-read and retry.
 Large files: tilth_read shows outline — use section to get hashlined content.
-After editing, tilth_edit returns a compact diff (old/new lines) followed by hashlined context.
+Pass diff: true to see a compact before/after diff in the response.
 After editing a function signature, tilth_edit shows callers that may need updating.
 DO NOT use the host Edit tool. Use tilth_edit for all edits.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -47,5 +47,6 @@ Range: {"start": "<line>:<hash>", "end": "<line>:<hash>", "content": "..."}
 Delete: {"start": "<line>:<hash>", "content": ""}
 Hash mismatch → file changed, re-read and retry.
 Large files: tilth_read shows outline — use section to get hashlined content.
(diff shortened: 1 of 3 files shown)


jahala:
Looks great, thank you for the PR @QuangThai !! merged

PR #51, opened 2026-04-02
docs: tilth_diff design document

jahala:
## Summary

Design document for `tilth_diff` — AST-aware structural diff tool (issue #45).

- Full MCP tool definition with progressive disclosure (overview → file → function → expand)
- Output format examples for all modes (overview, file detail, function detail, merge conflicts, log)
- Data structures, pipeline, reuse matrix (12 existing modules reused)
- Feature decision matrix: 20 features DOING (Phase 1 + 2), 14 NOT DOING with rationale
- ~880 lines estimated, zero new dependencies

**Looking for feedback from @rnett and @QuangThai on the design before implementation begins.**

See `docs/ideas/tilth_diff_design.md` for the full document.

Refs #45

### DIFF
--- a/docs/ideas/tilth_diff_design.md
+++ b/docs/ideas/tilth_diff_design.md
@@ -1,0 +1,439 @@
+# tilth_diff — Design Document
+
+## The problem
+
+Agent does `git diff` on a 50-file PR → gets 3000+ lines of raw unified diff → burns 10k+ tokens reading it → still can't answer "what actually changed?" without multiple follow-up reads and searches.
+
+## What tilth_diff does
+
+One tool, progressive disclosure. Same pattern as tilth_read (full vs outline vs section) and tilth_search (overview → drill → expand).
+
+```
+tilth_diff()                                    # overview: file list + function-level markers
+tilth_diff(scope: "auth.rs")                    # one file: all changes with structural context
+tilth_diff(scope: "auth.rs:handleAuth")         # one function: before/after with line detail
+tilth_diff(search: "error")                     # search within changed code only
+tilth_diff(blast: true)                         # overview + downstream impact per symbol
+tilth_diff(patch: "fix.patch")                  # parse a patch file
+tilth_diff(a: "old.rs", b: "new.rs")            # file-to-file comparison
+tilth_diff(log: "HEAD~5..HEAD")                 # per-commit structural summaries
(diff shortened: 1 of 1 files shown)


QuangThai:
Hi @jahala 

I like the overall direction here: keeping `tilth_diff` as a progressive-disclosure structural overlay on top of git diff is the right product choice for Tilth.

I do think the current design should tighten the matching core before implementation starts.

The main concern is symbol identity. Right now the design matches old/new symbols by name and also detects cross-file moves by name alone. That will break on overloaded methods, impl blocks, duplicate helper names, nested methods, and rename+move refactors. I’d strongly recommend introducing a more stable symbol identity for the overlay layer, something like `(kind, parent_path, name, normalized_signature)`, then falling back to a lightweight body similarity heuristic only when exact identity fails. If a match is still ambiguous, surface that ambiguity instead of confidently emitting a move marker.

Related to that, I would not directly reuse the current `blast_radius()/touched_symbols()` model for signature changes. The existing blast logic is based on a fixed “first ~4 lines of the function” heuristic, which is fine for edit-mode blast checks but too weak for `tilth_diff` where “signature changed” is a first-class concept. It would be more robust to compute `signature_changed` from matched old/new normalized signatures, then run caller analysis only for those confirmed symbols.

I’d also change file-to-file mode to use `git diff --no-index --no-color -U3 a b` instead of `diff -u`. The current doc says `diff -u` is “POSIX, always available”, which isn’t true on Windows, while this design already assumes git is present.

My suggested implementation shape would be:

1. Git metadata layer:
   - use git to get file status, rename/copy metadata, and hunks
2. Symbol overlay layer:
   - parse old/new trees with tree-sitter
   - build symbol tables with stable IDs
   - classify added/deleted/body-changed/signature-changed
   - only do move/rename inference when confidence is high
3. Formatter layer:
   - overview, file detail, function detail
4. Phase 2:
   - blast
   - search within diff
   - conflict mode
   - log mode
   - patch mode

That keeps the v1 kernel smaller and makes the hard part correct before adding more source modes.

Really looking forward to seeing this implemented and getting to try it out.

Thank you

rnett:
I came across https://github.com/Ataraxy-Labs/sem today, which looks like a very similar tool.  Might be worth reviewing it or even using it, since it's available as a library.

jahala:
Covered in #55

PR #54, opened 2026-04-03
refactor: tilth_run v2 — structured API, CLI, hook integration

jahala:
## Summary

Adds `tilth_run` — a command execution + output compression engine for the tilth MCP server and CLI. Parses verbose CLI output (cargo, npm, pytest, jest, eslint, go test, mypy, ruff, tsc) into structured diagnostics with never-worse compression guarantee.

### What's included

- **Structured API**: `process_structured()` returns `CompressResult` with lazy formatting (Markdown, Plain, JSON)
- **Command execution** (`exec.rs`): Process group killing via `setpgid(0,0)` + `killpg()`, bounded reads, lossy UTF-8, timeout handling
- **9 parsers**: cargo build/test, npm, jest, eslint, pytest, go test, mypy, ruff, tsc — each with AST-aware diagnostic extraction
- **CLI subcommand**: `tilth run <command>` with `--format`, `--timeout`, `--cwd` options
- **Hook integration**: `tilth hook-rewrite` for Claude Code PreToolUse hooks — transparently compresses Bash tool output
- **3 output formats**: Markdown (MCP), Plain (terminal/hook), JSON (machine consumption)

### Safety & correctness (3-round audit, 250 tests)

- Process group killing ensures entire command trees die on timeout (not just `sh`)
- `killpg(0)` guard prevents killing the MCP server's own process group
- Lossy UTF-8 reads — no panics on binary command output
- POSIX shell quoting with `'\''` escaping for hook paths
- Single-arg passthrough avoids re-quoting compound commands from hooks
- `extract_base_command()` skips env vars and wrappers (sudo, env, time, etc.)
- Never-worse guarantee: compressed output is always <= original size

## Test plan

- [x] 250 unit + integration tests pass (`cargo test`)
- [x] `cargo clippy -- -D warnings` clean
- [x] `cargo fmt --check` clean
- [x] 3 full audit rounds with test-first bug fixing
- [ ] Manual test: `tilth run "cargo build"` on a real project
- [ ] Manual test: hook rewrite integration with Claude Code

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -584,6 +584,7 @@ dependencies = [
  "memchr",
  "memmap2",
  "rayon",
+ "regex",
  "regex-syntax",
  "serde",
  "serde_json",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -581,6 +581,7 @@ dependencies = [
  "grep-regex",
  "grep-searcher",
  "ignore",
+ "libc",
  "memchr",
  "memmap2",
  "rayon",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,7 @@ regex-syntax = "0.8"
 
 # Byte scanning (SIMD-accelerated)
 memchr = "2"
+regex = "1"
 memmap2 = "0.9"
 
 # Parallelism
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,6 +63,8 @@ toml = "0.8"
 # MCP protocol (JSON-RPC over stdio)
 # (handled manually — no framework needed)
 
+[target.'cfg(unix)'.dependencies]
+libc = "0.2"
 
 [profile.release]
 opt-level = 3
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -24,11 +24,11 @@ pub mod install;
 pub mod map;
 pub mod mcp;
 pub(crate) mod read;
+#[allow(dead_code, clippy::pedantic)]
+pub(crate) mod run;
 pub(crate) mod search;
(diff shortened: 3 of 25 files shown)


sting8k:
Didn't deep-dive into this feature, but what I understand (after overview) is that it quite similiar to this: [RTK](https://github.com/rtk-ai/rtk)
Did I get it right?

jahala:
Hey @sting8k - I wasn't aware of RTK, but now that I am - I'm not sure tilth_run is really worth adding. I'll let it "hang" around in this PR for a while longer. Maybe someone else has a great idea about how to implement it in a way that is genuinely helpful.

sting8k:
There will be many confuse if we add this feat.
+ Agents have to route all their bash calls via this func? So it should implemented as a hook.
+ This is like a different project inside tilth. And I think (maybe) we redo what Rtk is doing well.

jahala:
Closing by the same standard we've applied to every other PR this week — this one's mine. The real-session measurement is unambiguous: across 126 real projects, build/test output through Bash is ~9% of tool-result bytes pooled and the median project is 0% — the heavy-build sessions were our own synthetic benchmark tasks. The demand this was built for doesn't show up in real usage, and the unanswered design critique (this only makes sense as a hook, not a standalone tool) was correct.

The branch and its stash stay preserved as a parts bin — the process-group execution semantics and the 12 per-tool output parsers are solid, reusable modules. Revival conditions, recorded: evidence of real build-output bloat in long sessions, plus a hook-based redesign. Neither is on the horizon, so this stops appearing as in-flight work.

PR #59, opened 2026-04-14
Add Elixir language support

dimitarvp:
🤖⚠️ ROBOT DISCLAIMER 🤖⚠️

Done with the help of Claude (max thinking effort). Curated and supervised by me, but not too heavily. I tried my best to guide it well and make sure there are no regressions -- but I make no promises that the code changes are not changing architecture in a way that the authors of this project would disapprove of. Receptive to criticisms on architecture and would iterate on them -- I would love for Tilth to have support for Elixir. If all goes well, I could make a similar PR for adding Erlang support.

## Summary

Adds Elixir language support: outline extraction, symbol search (definitions + usages), callee/caller detection, and import recognition. Uses `tree-sitter-elixir` v0.3. Both `.ex` and `.exs` files are supported.

Supported constructs: `defmodule`, `def`/`defp`, `defmacro`/`defmacrop`, `defguard`/`defguardp`, `defdelegate`, `defstruct`, `defexception`, `defprotocol`, `defimpl`, `@type`/`@typep`/`@opaque`, `@callback`/`@macrocallback`. Guard clauses (`when`), multi-clause functions, no-arg functions, nested modules, and pipe operator callees all work.

## Implementation notes

Elixir's tree-sitter grammar is structurally different from every other supported language. In Rust/Go/Python/etc., definitions have distinct node kinds (`function_declaration`, `struct_item`, `class_definition`). In Elixir, **everything is a `call` node** — `defmodule`, `def`, `defstruct` are all calls whose `target` identifier happens to be a definition keyword.

This means `DEFINITION_KINDS` (the shared list of definition node kinds) cannot include `"call"` without breaking all other languages. Instead, dedicated helpers check whether a `call` node's target is an Elixir definition keyword:

- `is_elixir_definition(node, lines)` — checks `call.target` against `defmodule`, `def`, `defp`, etc.
- `extract_elixir_definition_name(node, lines)` — navigates from `call` → `arguments` → first child to extract the defined name
- `elixir_extract_func_head_name(node, lines)` — handles the three shapes a function head can take: `call` (normal), `identifier` (no-arg), or `binary_operator` (guard clause with `when`)

A second grammar quirk: `arguments` is a node **kind**, not a named **field**. `child_by_field_name("arguments")` returns `None` — you have to find it by walking children and matching the kind. The `elixir_arguments()` helper handles this.

## Files changed

| File | What |
|------|------|
| `Cargo.toml` | Added `tree-sitter-elixir = "0.3"` |
| `src/types.rs` | `Lang::Elixir` variant |
| `src/lang/mod.rs` | `.ex`/`.exs` detection, `mix.exs` in package root manifests |
| `src/lang/outline.rs` | Outline extraction for all Elixir definition forms, `@type`/`@callback` attributes, `@doc`/`@moduledoc` extraction, `do`-block signature truncation |
| `src/lang/treesitter.rs` | `elixir_arguments()`, `is_elixir_definition()`, `extract_elixir_definition_name()`, `elixir_extract_func_head_name()`, `elixir_definition_weight()` |
| `src/index/symbol.rs` | Symbol index recognizes Elixir call-node definitions |
| `src/search/symbol.rs` | Symbol search finds Elixir definitions |
| `src/search/callees.rs` | Callee query for local and remote/dot calls |
| `src/search/callers.rs` | `find_enclosing_function` recognizes Elixir definitions as containers |
| `src/read/imports.rs` | `alias`, `import`, `use`, `require` detection |
| `src/overview.rs` | Display name |

## Test plan

8 new tests covering: dotted module names, `def`/`defp`/`defmacro` (block and keyword forms), guard clauses with `when`, `defguard`, multi-clause functions, no-arg functions, `defdelegate`, `defprotocol`/`defimpl`, `defexception`, nested modules, local calls, remote/dot calls (`Enum.map`), and pipe operator callees.

- [x] `cargo test` — all pass, no regressions
- [x] Verified against a real Elixir project — structured outlines, symbol definitions, and callers work on both `.ex` and `.exs` files
- [x] Side-by-side comparison with Rust and Go output — Elixir on par across all dimensions

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -712,6 +712,7 @@ dependencies = [
  "tree-sitter-c",
  "tree-sitter-c-sharp",
  "tree-sitter-cpp",
+ "tree-sitter-elixir",
  "tree-sitter-go",
  "tree-sitter-java",
  "tree-sitter-javascript",
@@ -805,6 +806,16 @@ name = "tree-sitter-cpp"
 version = "0.23.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "df2196ea9d47b4ab4a31b9297eaa5a5d19a0b121dceb9f118f6790ad0ab94743"
+dependencies = [
+ "cc",
+ "tree-sitter-language",
+]
+
+[[package]]
+name = "tree-sitter-elixir"
+version = "0.3.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "66dd064a762ed95bfc29857fa3cb7403bb1e5cb88112de0f6341b7e47284ba40"
 dependencies = [
  "cc",
  "tree-sitter-language",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ tree-sitter-scala = "0.24"
 tree-sitter-c-sharp = "0.23"
 tree-sitter-swift = "0.7"
 tree-sitter-kotlin-ng = "1.1"
+tree-sitter-elixir = "0.3"
 
 # Search (ripgrep internals)
 grep-regex = "0.1"
--- a/src/index/symbol.rs
+++ b/src/index/symbol.rs
@@ -310,6 +310,12 @@ fn walk_definitions(
                 symbols.push((Arc::from(iface.as_str()), line, true));
             }
         }
+    } else if crate::lang::treesitter::is_elixir_definition(node, lines) {
+        // Elixir: all definitions are `call` nodes — handle separately
+        if let Some(name) = crate::lang::treesitter::extract_elixir_definition_name(node, lines) {
+            let line = node.start_position().row as u32 + 1;
+            symbols.push((Arc::from(name.as_str()), line, true));
+        }
     }
 
     // Recurse into children for nested definitions (impl blocks, class bodies, modules)
--- a/src/index/symbol.rs
+++ b/src/index/symbol.rs
@@ -260,7 +260,7 @@ fn extract_symbols(path: &Path, content: &str) -> Vec<(Arc<str>, u32, bool)> {
     let lines: Vec<&str> = content.lines().collect();
     let mut symbols = Vec::new();
 
-    walk_definitions(tree.root_node(), &lines, &mut symbols, 0);
+    walk_definitions(tree.root_node(), &lines, &mut symbols, lang, 0);
 
     symbols
 }
@@ -275,6 +275,7 @@ fn walk_definitions(
     node: tree_sitter::Node,
     lines: &[&str],
     symbols: &mut Vec<(Arc<str>, u32, bool)>,
+    lang: crate::types::Lang,
     depth: usize,
 ) {
     if depth > 3 {
@@ -310,7 +311,9 @@ fn walk_definitions(
                 symbols.push((Arc::from(iface.as_str()), line, true));
             }
         }
-    } else if crate::lang::treesitter::is_elixir_definition(node, lines) {
+    } else if lang == crate::types::Lang::Elixir
+        && crate::lang::treesitter::is_elixir_definition(node, lines)
+    {
         // Elixir: all definitions are `call` nodes — handle separately
         if let Some(name) = crate::lang::treesitter::extract_elixir_definition_name(node, lines) {
             let line = node.start_position().row as u32 + 1;
@@ -321,7 +324,7 @@ fn walk_definitions(
     // Recurse into children for nested definitions (impl blocks, class bodies, modules)
     let mut cursor = node.walk();
     for child in node.children(&mut cursor) {
-        walk_definitions(child, lines, symbols, depth + 1);
+        walk_definitions(child, lines, symbols, lang, depth + 1);
     }
 }
 
--- a/src/lang/mod.rs
+++ b/src/lang/mod.rs
@@ -24,6 +24,7 @@ pub fn detect_file_type(path: &Path) -> FileType {
         Some("swift") => FileType::Code(Lang::Swift),
         Some("kt" | "kts") => FileType::Code(Lang::Kotlin),
         Some("cs") => FileType::Code(Lang::CSharp),
+        Some("ex" | "exs") => FileType::Code(Lang::Elixir),
 
         Some("md" | "mdx" | "rst") => FileType::Markdown,
         Some("json" | "yaml" | "yml" | "toml" | "xml" | "ini") => FileType::StructuredData,
@@ -56,6 +57,7 @@ pub(crate) fn package_root(path: &Path) -> Option<&Path> {
         "pom.xml",
         "build.gradle",
         "build.sbt",
+        "mix.exs",
     ];
     let mut dir = path;
     loop {
--- a/src/lang/outline.rs
+++ b/src/lang/outline.rs
@@ -514,32 +514,9 @@ fn elixir_func_name(node: tree_sitter::Node, lines: &[&str]) -> Option<String> {
         if !child.is_named() {
             continue;
         }
-        return elixir_func_head_name(child, lines);
+        return super::treesitter::elixir_extract_func_head_name(child, lines);
     }
     None
-}
-
-/// Extract function name from a function head node.
-///
-/// Handles three shapes:
-/// - `call`: `greet(name)` → target is `greet`
-/// - `identifier`: `bar` (no-arg, no-parens)
-/// - `binary_operator[when]`: `foo(x) when x > 0` → unwrap left, recurse
-fn elixir_func_head_name(node: tree_sitter::Node, lines: &[&str]) -> Option<String> {
-    match node.kind() {
-        "call" => {
-            if let Some(target) = node.child_by_field_name("target") {
-                return Some(node_text(target, lines));
-            }
-            None
-        }
-        "identifier" => Some(node_text(node, lines)),
-        "binary_operator" => {
-            let left = node.child_by_field_name("left")?;
-            elixir_func_head_name(left, lines)
-        }
-        _ => None,
-    }
 }
 
 /// Extract type name from an Elixir `@type` call.
@@ -586,11 +563,11 @@ fn elixir_callback_name(call: tree_sitter::Node, lines: &[&str]) -> Option<Strin
         if child.kind() == "binary_operator" {
             // `handle_event(...) :: return_type` → left is the function head
             if let Some(left) = child.child_by_field_name("left") {
-                return elixir_func_head_name(left, lines);
+                return super::treesitter::elixir_extract_func_head_name(left, lines);
             }
         }
         // Bare callback without return type spec (unlikely but handle it)
-        return elixir_func_head_name(child, lines);
+        return super::treesitter::elixir_extract_func_head_name(child, lines);
     }
     None
 }
--- a/src/lang/outline.rs
+++ b/src/lang/outline.rs
@@ -375,7 +375,11 @@ fn extract_doc(node: tree_sitter::Node, lines: &[&str]) -> Option<String> {
 // Elixir-specific outline helpers
 // ---------------------------------------------------------------------------
 
-/// Elixir definition keywords that correspond to outline entries.
+/// Elixir function-like definition keywords that produce `OutlineKind::Function`.
+/// This is the subset of definition keywords handled uniformly (extract function
+/// name from arguments). Container keywords (`defmodule`, `defprotocol`, `defimpl`,
+/// `defstruct`, `defexception`) have their own match arms in `elixir_call_to_entry`.
+/// See also `ELIXIR_DEFINITION_TARGETS` in `treesitter.rs` for the complete set.
 const ELIXIR_DEF_KEYWORDS: &[&str] = &[
     "def",
     "defp",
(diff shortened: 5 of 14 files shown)


jahala:
Hey @dimitarvp - thank you so much for the PR!

I had Opus 4.6 review it -- please see the output below

---

Thanks for this — really well-structured PR. The Elixir grammar choice is solid (official `elixir-lang` org, 1.5M downloads, Apache-2.0, used by Helix/Neovim/Zed), and the approach to handling Elixir's "everything is a `call` node" grammar quirk is well thought through.

I went through the code carefully. Two things I'd like addressed before merging, plus a number of smaller items worth discussing or noting for follow-up.

---

## Must fix

### 1. `is_elixir_definition` needs a language guard or documented safety argument

In `walk_for_definitions` (symbol.rs) and `walk_definitions` (index/symbol.rs), the `else if is_elixir_definition(node, lines)` branch runs unconditionally on every file in every language. It happens to be safe today because no other supported grammar uses `call` nodes with a `target` field — Ruby uses `method`, Python uses `function`, etc. But this is an undocumented accident, and if any future grammar introduces a `call` node with a `target` field, this silently misfires.

Either:
- Thread `lang` through both walk functions and guard with `lang == Lang::Elixir`, or
- Add a comment documenting which grammars were verified and why the `target` field discriminates safely

I'd prefer the explicit lang guard — it's a small change and makes the contract clear.

### 2. Add explicit `Lang::Elixir` arm to `is_external` in `read/imports.rs`

Currently falls through to `_ => true`, which is correct behavior (no Elixir cross-file resolution exists), but it should be an explicit arm rather than an invisible catch-all:

```rust
Lang::Elixir => true, // no cross-file resolution yet
```

Same applies to `resolve()` if it has a similar catch-all.

---

## Should fix

### 3. Duplicate `elixir_func_head_name` implementations

`outline.rs:elixir_func_head_name` and `treesitter.rs:elixir_extract_func_head_name` are structurally identical — same match arms, same logic, same recursion for `binary_operator`. Could you extract one and reuse it? Having two copies is a maintenance hazard that will diverge over time.

### 4. Callee query captures definition keywords as callees

The query `(call target: (identifier) @callee)` also captures `def`, `defmodule`, `defp` etc. as callee names since those are all `call` nodes with identifier targets. Your tests verify that real callees (`query`, `map`, `puts`, `local_func`) are found, but don't assert that definition keywords are absent from the results. Worth adding a negative assertion like:

```rust
assert!(!names.contains(&"def".to_string()), "definition keywords should not appear as callees");
assert!(!names.contains(&"defmodule".to_string()), "...");
```

If they do appear, the callee query needs a filter.

### 5. Elixir `use` handled by Rust branch in `extract_import_source`

`use GenServer` matches the Rust `use ` prefix strip before reaching your new Elixir block (which only covers `alias` and `require`). The end result is accidentally correct — returns `"GenServer"`, `is_external` returns `true`, no resolution attempted. But the Elixir code path is never actually reached for `use` statements.

If we later add Elixir file resolution, this will silently misroute `use` lines through Rust's module path logic. Either:
- Add `"use "` to your Elixir prefix list and handle the ordering (with a lang check), or
- Add a comment noting the accidental ordering dependency so future work doesn't miss it

### 6. Partially overlapping keyword constants

`ELIXIR_DEF_KEYWORDS` in `outline.rs` and `ELIXIR_DEFINITION_TARGETS` in `treesitter.rs` partially overlap but are not identical. The split is intentional (outline handles container keywords like `defstruct`/`defprotocol`/`defimpl` separately), but this isn't documented anywhere. A comment on each constant pointing to the other would prevent future maintainers from thinking one list is the canonical source of truth.

---

## Worth discussing

### 7. `defstruct`/`defexception` indexed as literal keyword strings

`extract_elixir_definition_name` returns `"defstruct"` for `defstruct [:a, :b]`. This means:
- Searching for `defstruct` returns every struct in the codebase
- Searching for the actual struct by its module name won't find the `defstruct` definition specifically

In Elixir the struct is conceptually the module (`%MyModule{}`), so the struct "name" is the enclosing `defmodule`. Not a blocker — just want to confirm this is the intended tradeoff. A comment explaining why would help.

### 8. `elixir_extract_doc` fragile string trimming

```rust
let trimmed = text
    .trim_matches('"')
    .trim_start_matches("~S\"\"\"")
    .trim_end_matches("\"\"\"")
    .trim();
```

This doesn't handle `~s"""` (lowercase sigil), heredocs with leading indentation, `@doc false`, or escaped quotes inside strings (`@doc "say \"hi\""` → broken output). Acceptable for best-effort outline doc extraction, but worth a comment noting the known limitations so future contributors don't assume it's comprehensive.

### 9. Module child outline includes import entries

`elixir_collect_children` walks `do_block` children and emits everything `node_to_entry` can handle, including `use`, `alias`, `import`, `require` as `OutlineKind::Import` entries inside module outlines. Other languages suppress imports in outline children. Whether this is desirable for Elixir is a product decision — `use GenServer` inside a module is arguably structural context, not just an import. Worth a brief comment on the intentional choice either way.

### 10. Elixir module qualification in callers bypasses `TYPE_KINDS`

In `find_enclosing_function` (callers.rs), all other languages qualify function names via the `TYPE_KINDS` constant. The PR adds a separate Elixir-specific check after the `TYPE_KINDS` walk because `defmodule` is a `call` node, not a type kind. This is correct behavior but creates an asymmetric code path that should have a comment explaining why.

---

## Nitpicks

- **Blank line removal in symbol.rs**: The diff removes a blank line between `let kind = node.kind();` and `if DEFINITION_KINDS.contains(&kind)`. Cosmetic, but was it intentional?
- **`elixir_arguments` cursor pattern**: The cursor created in `elixir_arguments()` is dropped while the returned `Node` survives — works because tree-sitter nodes are `Copy` (arena indices), but a brief comment would help future maintainers unfamiliar with tree-sitter's lifetime model.

---

## What looks good

- Test coverage is solid — 8 tests covering guard clauses, multi-clause, no-arg, protocols, pipes, nested modules, delegates, exceptions
- The `tree-sitter-elixir` crate is excellent quality (official `elixir-lang` org, 277 stars, 2 open issues, actively maintained, Apache-2.0)
- `mix.exs` in package root manifests is correct
- Import detection for `alias`/`import`/`use`/`require` is complete
- The PR description and implementation notes are thorough — especially the explanation of why `DEFINITION_KINDS` can't include `"call"`

Looking forward to iterating on this. Would love to have Elixir support in tilth.

dimitarvp:
Thank you for following up! Me & Claude are both on all of these and they'll be addressed.

dimitarvp:
This comment has been posted by Claude with the permission of @dimitarvp.

---

All 12 review items addressed in 6 commits:

**Must fix:**
- `3431289` — Language guard: threaded `lang` through `walk_for_definitions`, `walk_definitions`, and `find_enclosing_function`. All Elixir-specific checks now guarded with `lang == Lang::Elixir`. Added explicit Elixir mention to `is_external`/`resolve` wildcard comments (separate arms trigger clippy `match-same-arms`).

**Should fix:**
- `d312110` — Deduplicated: deleted `outline.rs::elixir_func_head_name`, shared version from `treesitter.rs` is now `pub(crate)`.
- `24d8654` — Callee filtering: post-filter in `extract_callee_names` strips all definition + import keywords for Elixir. Negative assertions added to tests.
- `7e25c3e` — Import source routing: added `lang` parameter to `extract_import_source`, Elixir branch runs first (before Rust `use` / JS `import` can misroute). All 4 callers updated.
- Cross-reference comments added to both `ELIXIR_DEF_KEYWORDS` and `ELIXIR_DEFINITION_TARGETS` explaining the intentional split.

**Worth discussing → resolved:**
- `defstruct`/`defexception` naming: documented that Elixir structs are their enclosing module (one per module, no standalone name).
- `d20755b` — Doc extraction rewritten: tree-sitter node type dispatch (`boolean` → None for `@doc false`, `string`/`sigil` → extract first content line) replaces fragile string trimming.
- Import entries in module outlines: documented as intentional (Elixir's `use`/`alias` are structural).
- TYPE_KINDS bypass in callers: documented why `defmodule` needs a separate check.

**Nitpicks:**
- Blank line restored.
- Cursor lifetime comment added to `elixir_arguments`.

jahala:
Hey again @dimitarvp 

Here is the new review

-----

Thanks for the thorough follow-up — all 12 original review items have been addressed, and the code is substantially cleaner. The atomic commit structure (fix, refactor, docs) made it easy to verify each change.

Quick re-review of what was fixed and a few remaining items I spotted.

---

## Original findings — all addressed

| # | Finding | Status |
|---|---------|--------|
| 1 | `is_elixir_definition` lang guard | **Fixed** — `lang` threaded through `walk_for_definitions` (as `Option<Lang>`) and `walk_definitions` (as `Lang`), both guarded with `lang == Elixir` |
| 2 | `is_external` explicit Elixir arm | **Fixed** — catch-all comment updated to name Elixir explicitly, `resolve()` likewise |
| 3 | Duplicate `elixir_func_head_name` | **Fixed** — single `pub(crate)` impl in `treesitter.rs`, outline.rs delegates via `super::treesitter::elixir_extract_func_head_name` |
| 4 | Callee keyword filtering | **Fixed** — `is_elixir_keyword` filter + negative assertions for `def`/`defmodule` in test |
| 5 | Elixir `use` consumed by Rust branch | **Fixed** — Elixir-first branch gated on `lang == Some(Lang::Elixir)` at top of `extract_import_source` |
| 6 | Keyword constant cross-references | **Fixed** — both `ELIXIR_DEF_KEYWORDS` and `ELIXIR_DEFINITION_TARGETS` now point to each other |
| 7 | `defstruct`/`defexception` as keyword | **Fixed** — documented with rationale ("struct IS its enclosing module") |
| 8 | `elixir_extract_doc` fragile trimming | **Fixed** — rewritten to use tree-sitter node types (`string`/`sigil`/`boolean`), handles `@doc false`, heredocs via interior line scanning |
| 9 | Module child outline includes imports | **Fixed** — `elixir_collect_children` now has a doc comment explaining the intentional choice |
| 10 | Callers bypass `TYPE_KINDS` | **Fixed** — commented explaining why `defmodule` needs a separate check |
| 11 | Blank line removal | Not present in v2 diff |
| 12 | `elixir_arguments` cursor lifetime | **Fixed** — comment added |

Signature changes to `extract_import_source`, `find_enclosing_function`, `find_defs_treesitter`, and `walk_for_definitions` are all propagated correctly to every call site. I verified all 4 callers of `extract_import_source`, both callers of `find_enclosing_function` in `callers.rs` (the one in `diff/overlay.rs` is a different function with the same name — takes `&[OutlineEntry]`, not tree-sitter nodes), and all 4 callers of `find_defs_treesitter`.

---

## New items found in v2

### Should fix before merge

**1. `defimpl` test is tautological (doesn't actually test defimpl)**

In `elixir_protocol_impl_exception`, the defimpl assertion:

```rust
// defimpl
assert!(
    !elixir_find(code, "Printable").is_empty(),
    "should find defimpl for 'Printable'"
);
```

...is identical to the protocol assertion two lines above. It passes because `defprotocol Printable` already matches — the defimpl match is never distinguished.

More importantly: `defimpl Printable, for: User` is indexed under `"Printable"` (the protocol name), not `"User"` (the implementing type). This means `tilth_search("User")` won't find the impl. This is because `extract_elixir_definition_name` routes `"defimpl"` through the same `"defmodule" | "defprotocol" | "defimpl"` arm, which takes the first named argument.

Two things needed:
1. Fix the test to assert something meaningful — e.g., verify that searching for `"Printable"` returns at least 2 results (protocol + impl), or count them explicitly
2. Add a comment in `extract_elixir_definition_name` noting that `defimpl` is indexed under the protocol name, so searching for the `for:` target won't find the impl

**2. `is_elixir_keyword` in `callees.rs` is a third keyword list with no cross-reference**

`ELIXIR_DEF_KEYWORDS` (outline.rs) and `ELIXIR_DEFINITION_TARGETS` (treesitter.rs) now cross-reference each other. But `is_elixir_keyword` in `callees.rs` is a third list that is a superset of both (adds `defoverridable`, `use`, `import`, `alias`, `require`). It has no cross-reference comment and will diverge if someone adds a new definition keyword to only one of the three places. A one-line comment would prevent this.

### Minor 

**3. ` do` truncation in `extract_signature` runs for all languages**

The new block:

```rust
if let Some(pos) = line.rfind(" do") {
    let after = &line[pos + 3..];
    if after.is_empty() || after.starts_with('\n') {
        return line[..pos].trim().to_string();
    }
}
```

This is inside `extract_signature` which is called from `node_to_entry` for every language. It's practically safe today — C/Java/Go/Rust definitions always hit the `{` branch first, Python hits the `:` branch — but if Ruby outline support is ever added, Ruby's `do...end` blocks would be truncated. A comment like `// Elixir: truncate at block keyword` would clarify the intent.

**4. `elixir_extract_doc_string` single-line sigil edge case**

For `@doc ~s"text"`, `start_row == end_row` is true, so the single-line branch fires: `text.trim_matches('"')` on `~s"text"`. The `~s` prefix survives because `trim_matches` only strips the outermost matching characters. This is an unlikely edge case (single-line sigils for `@doc` are rare), but it will produce `~s"text` in the outline doc. Not a blocker — just noting it.

---

## Verdict

This is ready to merge after items 1 and 2 above. Both are small fixes — one test tweak and one comment. Everything else is solid.

Really nice work adapting to a grammar that breaks every assumption the rest of the codebase relies on. The tree-sitter node type approach for doc extraction (v2) is much better than the string trimming (v1). Looking forward to merging this and potentially Erlang support next.

dimitarvp:
We are almost there. Pasting Claude's summary on how the follow-up feedback was addressed.

---

All 4 v2 items addressed in 3 commits:

- `7e120a5` — defimpl test: replaced tautological assertion with count check (`>= 2` for protocol + impl). Documented in `extract_elixir_definition_name` that defimpl is indexed under the protocol name, not the `for:` target.
- `4142b1f` — Single-line sigil doc: `@doc ~s"text"` now strips the `~s` prefix before trimming quotes.
- `dbe102b` — Cross-reference comment on `is_elixir_keyword` pointing to `ELIXIR_DEFINITION_TARGETS`. Comment on ` do` truncation explaining why it's safe for non-Elixir languages.

jahala:
Thank you so much for your contribution @dimitarvp! 🙏🏻

PR #60, opened 2026-04-19
feat(mcp): per-request wall-clock timeout with crossbeam-channel

paulnsorensen:
## Summary

Adds a per-request wall-clock timeout to all MCP tool calls, preventing slow or stuck operations from blocking the JSON-RPC loop indefinitely.

### What's included

- **`src/timeout.rs`** — new module with `spawn_with_timeout()`: spawns a worker thread, waits on a `crossbeam_channel` bounded(1) with `select! { default(timeout) => ... }`, and detaches the worker on deadline
- **`ThreadTracker`** — atomic counter of still-running abandoned threads; refuses new work at `MAX_ABANDONED_THREADS` (8) to prevent unbounded thread accumulation
- **`ThreadCoord`** — `RUNNING` → (`TIMED_OUT` | `FINISHED`) CAS state machine guaranteeing the tracker is incremented/decremented exactly once per spawn, even when the worker's send races the main thread's deadline
- **`Services` struct** in `src/mcp.rs` — consolidates shared dependencies (`OutlineCache`, `Session`, `SymbolIndex`, `BloomFilterCache`, `ThreadTracker`) into one `Clone`-able handle passed through the request pipeline
- **`TILTH_TIMEOUT` env var** — configures timeout in seconds (default: 30); error message instructs callers to raise it if needed
- **Warn threshold** at 3 abandoned threads, hard cap at 8

### Why crossbeam-channel

Rust has no sync-world equivalent of `Future.get(timeout, unit)`. `std::sync::mpsc` doesn't support `select!` with a deadline; `crossbeam_channel::select!` does, with no async executor overhead.

### Safety & correctness

- CAS protocol ensures tracker counter is updated at most once per spawn under all races
- Hard cap prevents thread accumulation on pathologically slow I/O (e.g. huge file, cold NFS mount)
- Timed-out workers run to natural completion — no forced cancellation, no undefined behaviour
- Env-var mutation in tests serialized via a `Mutex` to prevent parallel-test races

## Test plan

- [x] `cargo test` passes (including timeout unit tests)
- [x] `cargo clippy -- -D warnings` clean
- [x] `cargo fmt --check` clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #61, opened 2026-04-19
feat(mcp): request timeouts, batched tilth_edit, and parallel diff overlay

paulnsorensen:
## Summary

Three related changes that together make the MCP server faster under large workloads and keep the JSON-RPC loop responsive.

1. **Request timeouts** — wall-clock guard on every tool call via a sync `select!`-with-deadline primitive, with bounded thread accumulation.
2. **Batched `tilth_edit`** — takes `files: [{path, edits}]` instead of single-file, rayon-parallel per file, partial failures don't block siblings. **Breaking schema change → 0.7.0.**
3. **Parallel diff overlay** — `compute_overlay` is pure per-file work; switching to `par_iter()` fans out across cores.

### Request timeouts (new `src/timeout.rs`)

Wraps every MCP tool call in a wall-clock guard so slow or stuck operations cannot block the JSON-RPC loop.

- `spawn_with_timeout()` spawns a worker thread and waits on a `crossbeam_channel` bounded(1) with `select! { default(timeout) => ... }`, detaching the worker on deadline. `std::sync::mpsc` has no `select!`-with-deadline primitive; `crossbeam_channel::select!` is the sync-world equivalent of `Future.get(timeout, unit)` with no executor overhead.
- `ThreadTracker` — atomic counter of still-running abandoned threads; refuses new work at `MAX_ABANDONED_THREADS` (8) to prevent unbounded thread accumulation on pathologically slow I/O (huge files, cold NFS).
- `ThreadCoord` — `RUNNING` → (`TIMED_OUT` | `FINISHED`) CAS state machine that guarantees the tracker is incremented/decremented exactly once per spawn, even when the worker's send races the main thread's deadline.
- `Services` struct in `mcp.rs` consolidates shared dependencies (`OutlineCache`, `Session`, `SymbolIndex`, `BloomFilterCache`, `ThreadTracker`) into one `Clone`-able handle threaded through the request pipeline.
- `TILTH_TIMEOUT` env var configures timeout in seconds (default 30; 90 for `tilth_edit`). Error messages instruct callers to raise it if needed. Warn threshold at 3 abandoned threads, hard cap at 8.
- Timed-out workers run to natural completion — no forced cancellation, no undefined behaviour.

### Batch `tilth_edit` (breaking schema change, 0.7.0)

`tilth_edit` now takes `files: [{path, edits}]` instead of single `path` + `edits`. Each file is processed independently — a hash mismatch on one file does not block its siblings — and per-file results are reported under `## <path>` headers separated by `---`. Max 20 files per call.

- New `src/edit.rs::apply_batch` mirrors the existing `read_batch` pattern: pulls the per-file loop out of `mcp.rs` and rayon-parallelises independent file work (canonicalize + `apply_edits` + `blast_radius` are all pure per-file work).
- `tool_edit` shrinks from ~70 lines to ~30: parse JSON, cap at 20, delegate to `apply_batch`. JSON parsing stays at the MCP wire boundary; parallel I/O and blast radius live next to `apply_edits`.
- `parse_edit_entry` extracted from `parse_file_edit` (was 61 lines, triple-nested).
- `FileEditTask` is the shared type carried from the parser through `apply_batch`. Parse errors get pre-labelled and rendered uniformly with execution errors.
- Per-request timeout covers the whole batch — no separate `TILTH_BATCH_TIMEOUT` needed.

### Parallel diff overlay (rayon)

`compute_overlay` is pure per-`FileDiff` work: each file independently parses old/new sources, runs tree-sitter, and matches symbols. Switching the overlay build from `.iter()` to `.par_iter()` lets multi-file diffs fan out across cores — a meaningful win for large PRs where each file does non-trivial AST work. `FileDiff` and `DiffSource` are `Send+Sync`, so the change is mechanical.

### MCP instructions + AGENTS.md

- Lead with `ALWAYS BATCH` — multi-file reads and edits should be one call with `paths:` / `files:`, never N serial calls.
- Document the new `tilth_edit` `files: [...]` shape with examples and per-file partial-failure semantics.
- Schema: `paths` and `files` declare `minItems:1/maxItems:20` to match runtime enforcement.
- Sync `EDIT_MODE_EXTRA` and `AGENTS.md` (including the missing `tilth_diff` block in `AGENTS.md`).

## Test plan

- [x] `cargo test` — 249 tests pass (13 new: 5 batch edit, 8 timeout)
  - `batch_edit_two_files_succeeds`
  - `batch_edit_partial_failure_does_not_block_siblings`
  - `batch_edit_over_limit_rejected`
  - `batch_edit_all_failed_returns_err`
  - `batch_edit_empty_files_array_rejected`
  - `spawn_with_timeout` happy path, deadline, abandoned-thread cap, CAS race, `TILTH_TIMEOUT` env parsing, warn threshold, poisoned mutex recovery
- [x] Env-var mutation in tests serialised via a `Mutex` to prevent parallel-test races

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,5 +1,7 @@
 tilth — code intelligence MCP server. Replaces grep, cat, find, ls with AST-aware equivalents.
 
+ALWAYS BATCH. When you have 2+ files to read, call tilth_read once with paths: [...]. When you have edits to multiple files, call tilth_edit once with files: [...]. Never make N serial calls when one will do — each tool call is a turn.
+
 To explore code, always search first. tilth_search finds definitions, usages, and file locations in one call.
 Usage: tilth_search(query: "handleRequest").
 tilth_files is ONLY for listing directory contents when you have no symbol or text to search for.
@@ -22,10 +24,11 @@ Output per match:
 <name>  <path>:<start>-<end>  <signature>
 Re-expanding a previously shown definition returns [shown earlier].
 
-tilth_read: Read file content with smart outlining. Replaces cat/head/tail.
+tilth_read: Read one or more files with smart outlining. Replaces cat/head/tail.
+paths: ["a.rs", "b.rs"] reads many in one call (max 20). PREFER this over serial single-file reads.
+path: "a.rs" for a single file.
 Small files → full content. Large files → structural outline.
 section: "<start>-<end>" or "<heading text>"
-paths: read multiple files in one call.
 Output:
 <line_number> │ <content>                  ← full/section mode
 [<start>-<end>]  <symbol name>             ← outline mode
@@ -36,18 +39,31 @@ Output: <path>  (~<token_count> tokens). Respects .gitignore.
 tilth_deps: Blast-radius check — what imports this file and what it imports.
 Use ONLY before renaming, removing, or changing an export's signature.
 
+tilth_diff: Structural diff — shows what changed at function level. Replaces Bash(git diff).
+Usage: tilth_diff(source: "HEAD~1") for last commit. No args = uncommitted changes.
+scope: "file.rs" or "file.rs:fn_name". log: "HEAD~5..HEAD" for per-commit summaries.
+search: filter to lines matching a term. blast: true to show callers of changed signatures.
+Output: [+] added, [-] deleted, [~] body changed, [~:sig] signature changed.
+DO NOT use Bash(git diff) or Bash(git log --patch). Use tilth_diff instead.
+
 To search code, use tilth_search instead of Grep or Bash(grep/rg).
 To read files, use tilth_read instead of Read or Bash(cat).
 To find files, use tilth_files instead of Glob or Bash(find/ls).
+To check what changed, use tilth_diff instead of Bash(git diff/git log).
 DO NOT re-read files already shown in expanded search results.
 
-tilth_edit: Edit files using hash-anchored lines. Replaces the host Edit tool.
+tilth_edit: Batch edit one or more files using hash-anchored lines. Replaces the host Edit tool.
+ALWAYS group edits to multiple files into ONE tilth_edit call (max 20 files). Never call tilth_edit twice in a row.
 tilth_read → copy anchors (<line>:<hash>) → pass to tilth_edit.
-Single line: {"start": "<line>:<hash>", "content": "<new code>"}
-Range: {"start": "<line>:<hash>", "end": "<line>:<hash>", "content": "..."}
-Delete: {"start": "<line>:<hash>", "content": ""}
-Hash mismatch → file changed, re-read and retry.
+Shape: {"files": [{"path": "a.rs", "edits": [...]}, {"path": "b.rs", "edits": [...]}]}
+Single file: {"files": [{"path": "a.rs", "edits": [{"start": "<line>:<hash>", "content": "<new code>"}]}]}
+Edit forms inside `edits`:
+  Single line: {"start": "<line>:<hash>", "content": "<new code>"}
+  Range:       {"start": "<line>:<hash>", "end": "<line>:<hash>", "content": "..."}
+  Delete:      {"start": "<line>:<hash>", "content": ""}
+Per-file results: each file is processed independently. A hash mismatch on one file does NOT block the others.
+Hash mismatch → file changed, re-read THAT file and retry it (other files in the batch already applied).
 Large files: tilth_read shows outline — use section to get hashlined content.
-Pass diff: true to see a compact before/after diff in the response.
+Pass diff: true to see a compact before/after diff per file.
 After editing a function signature, tilth_edit shows callers that may need updating.
 DO NOT use the host Edit tool. Use tilth_edit for all edits.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -156,6 +156,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b05b61dc5112cbb17e4b6cd61790d9845d13888356391624cbe7e41efeac1e75"
 
 [[package]]
+name = "crossbeam-channel"
+version = "0.5.15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "82b8f8f868b36967f9606790d1903570de9ceaf870a7bf9fbbd3016d636a2cb2"
+dependencies = [
+ "crossbeam-utils",
+]
+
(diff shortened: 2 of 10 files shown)


paulnsorensen:
@jahala if you're cool with this, I opened #60 as precursor, but looks like claude batched the commit message up. I can fix if you're interested in this change or we can ignore that PR depending on how you'd prefer it.

jahala:
Hey @paulnsorensen - first of all, thank you so much for contributing! 

As I've written in other comments, I'm not really a dev - so Opus 4.7 is doing all the heavy lifting for me in terms of code and reviewing. Hope you can have a look at the comments below. Thanks!

---

Thanks for the contribution and for squashing — the clean history on this one is appreciated. I've reviewed the code in detail (same diff as #60, so closing that one as superseded).

There are some good ideas here — the `Services` struct consolidation and the timeout CAS state machine in particular — but this PR has blocking issues that need to be resolved before it can be merged. The full review is below.

---

## Must split

This PR bundles 5 independent features:

1. Timeout module (`timeout.rs`) with CAS thread coordination
2. `Services` struct consolidation (`mcp.rs` rewrite)
3. Batch `tilth_edit`
4. Rayon parallelization (read_batch + diff overlay)
5. Version bump to 0.7.0

These need to be separate PRs. They have different risk profiles, different review concerns, and if any one of them needs to be reverted, the others get dragged along. I'd suggest:

- **(a)** `Services` struct refactor — pure refactor, no behavior change, easy to review and merge first
- **(b)** Timeout module — depends on (a) for `Services.tracker`
- **(c)** Batch `tilth_edit` — can be reviewed independently
- **(d)** Rayon parallelization — smallest, lowest risk

### Remove version bump

Version bumps are a maintainer decision. Please don't include them in external PRs.

### Remove `tilth_diff` from AGENTS.md

We haven't decided to surface `tilth_diff` in AGENTS.md yet — the MCP instructions (`SERVER_INSTRUCTIONS` in `mcp.rs`) already include it for MCP clients, but the user-facing AGENTS.md copy is intentionally different. Please remove this addition.

---

## Critical — must fix

### 1. `tilth_edit` schema change — clean break is fine, but document why

The schema change from `{"path", "edits"}` to `{"files": [{path, edits}]}` is a clean break. Since MCP serves `tools/list` dynamically on every session, clients always get the current schema — there's no stale-cache problem. The LLM sees the new `inputSchema` on next connect and adapts.

So backward compatibility isn't strictly needed here. But the PR should explicitly call this out: "this is a schema break, here's why it's safe (dynamic tool discovery, no cached schemas)." Silent breaking changes, even safe ones, need to be documented so reviewers don't have to figure out the safety argument themselves.

### 2. `dispatch_tool` visibility downgraded from `pub(crate)` to `fn`

This is `pub(crate)` intentionally — it's used for testing and potential benchmark/integration harness access. Please keep the existing visibility.

### 3. `session.record_read` inside rayon parallel closure — Mutex contention

In `read_batch`, `session.record_read(path)` is called inside `par_iter().map(...)`. `Session::record_read` acquires `self.dir_hits.lock()` — a `Mutex<HashMap>`. With rayon dispatching multiple threads for a 20-file batch, all threads serialize on the same mutex.

Fix: do the parallel reads first, then call `session.record_read` sequentially on the collected paths.

### 4. Batch edit allows duplicate file paths — data corruption

`apply_batch` dispatches via `into_par_iter()` with no uniqueness check on file paths. If two entries target the same file, both rayon workers read the file, compute edits against the same content, and race to write. Last writer wins, first edit silently lost.

Fix: reject or deduplicate before dispatching.

---

## High — should fix

### 5. `TILTH_BATCH_TIMEOUT` removal is a behavior regression

The old code had a cooperative per-file deadline in `tool_read` that could stop between files and return partial results with a helpful message. The new code does all files in parallel — on a slow 20-file batch, the LLM gets either everything or a hard timeout error, never partial results. The old behavior was better for large batches on slow filesystems.

### 6. `is_at_cap` uses `Ordering::Relaxed` — stale reads on ARM64

The `is_at_cap()` load uses `Ordering::Relaxed`. On weakly-ordered architectures (every Apple Silicon Mac), this can return stale data. The CAS operations on `ThreadCoord` correctly use `AcqRel`, but the counter check should use at least `Ordering::Acquire`.

### 7. Increment-before-CAS window causes false rejections

In `spawn_with_timeout`, the timeout arm calls `record_timeout()` (fetch_add) *before* `claim_timeout()` (CAS). Between those two operations, a concurrent `is_at_cap()` sees an inflated count. Under the hard cap of 8, a request can be falsely rejected when the actual abandoned count is 7.

---

## Medium

- **`parse_file_edit` short-circuits on first bad edit** — if `edits[0]` is valid and `edits[1]` is malformed, the whole file becomes `ParseError` and `edits[0]` is silently discarded. Error message should indicate which edit index failed.
- **`request_timeout_reads_env` test uses `std::env::set_var`** — this is `unsafe` in Rust 1.81+ in multi-threaded contexts. The `ENV_LOCK` doesn't cover other tests that spawn threads. Consider a direct parameter instead.
- **`batch_edit_all_failed_returns_err` test** uses fixed-name paths in `std::env::temp_dir()` instead of `tempfile::tempdir()`.
- **Diff overlay `par_iter` change** — `compute_overlay` calls tree-sitter C bindings. While the Rust wrappers are `Send`, a comment confirming this was verified would be appropriate.

---

## What looks good

- **`ThreadCoord` CAS state machine** — the `RUNNING` → (`TIMED_OUT` | `FINISHED`) protocol correctly ensures at-most-once counter updates. The `catch_unwind` in the worker thread ensuring the tracker is always updated even on panics is a nice touch.
- **`Services` struct** — genuinely good refactor that eliminates 6-parameter threading through the request pipeline.
- **Batch edit tests** — solid coverage: partial failure, over-limit, all-failed, empty array, two-file success.
- **`apply_one` / `render_applied` split** — keeps the parallel closure trivial and the per-file logic testable in isolation.

---

## Summary

Split into separate PRs. Remove the version bump and AGENTS.md `tilth_diff` addition. Address the thread safety issues (Mutex contention, duplicate-path check, Ordering::Acquire). Happy to review the individual PRs once split — the `Services` refactor in particular is a quick merge.

Closing #60 as superseded by this one.

jahala:
Superseded by the three splits that landed on main today:

#82 — timeout extraction
#84 — batch tilth_edit
#83 — rayon overlay parallelization

Closing this since the substance has landed. Thanks for the original bundle @paulnsorensen!

PR #62, opened 2026-04-20
fix: prevent walker from traversing NFS mounts and cross-device boundaries

quanhavn:
## Problem

`tilth_search` was timing out frequently when NFS mounts (e.g., OrbStack) or other mounted volumes were accessible from the project directory.

## Root Causes

1. **Walker followed symlinks** into NFS/mounted volumes (`follow_links(true)`)
2. **No cross-device boundary detection** - walker would traverse into any mounted filesystem
3. **Missing PHP manifest** - `composer.json` not recognized, causing fallback to incorrect scope
4. **No git fallback** - projects without language-specific manifests weren't detected

### `src/search/mod.rs`
- Set `follow_links(false)` to avoid following symlinks into mounted volumes
- Add `same_file_system(true)` to stop traversal at mount point boundaries

### `src/lang/mod.rs`
- Add `composer.json` to manifest list for PHP/Laravel projects
- Add `.git` as fallback manifest to detect any git repository root

## Testing

Tested on macOS with OrbStack NFS mount at `~/OrbStack`. Search no longer traverses into NFS mount and completes quickly.


jahala:
Hi @quanhavn - thank you so much for your contribution! 🙏🏻

Please see the review below by Opus 4.6: 

---

Thanks for the PR — NFS traversal causing timeouts is a real problem worth fixing. However, the proposed changes have significant side effects that would break common project layouts. Let me walk through each change.

---

## `follow_links(false)` — this breaks monorepos and linked packages

Disabling symlink following is too aggressive. Many real project layouts depend on symlinks:

- **pnpm** uses symlinked `node_modules` — every package in a pnpm workspace has symlinks to the shared store. With `follow_links(false)`, tilth can't find any installed dependency source.
- **Yarn/npm workspaces** with `link:` dependencies create symlinks between workspace packages.
- **Go modules** in `vendor/` can be symlinked.
- **Monorepos** with shared code symlinked across packages.
- **macOS** itself: `/tmp` is a symlink to `/private/tmp`. Any project in `/tmp` would be invisible.

The NFS problem is real, but the fix should be targeted at NFS specifically, not all symlinks. `same_file_system(true)` already handles the NFS/cross-device case — that's the correct fix by itself. Symlinks within the same filesystem (which covers all the above cases) are fine to follow.

**Recommendation:** Keep `follow_links(true)`, add `same_file_system(true)`. That stops NFS traversal without breaking symlinked packages.

## `same_file_system(true)` — this is the right fix

This prevents the walker from crossing mount point boundaries, which is exactly the NFS problem. On its own, this is correct and sufficient. Keep this change.

One caveat worth noting: on Docker/container setups, bind mounts from the host appear as the same filesystem. That's actually fine — you want to search those. The issue is NFS/network mounts which are always different devices.

## `.git` as fallback manifest in `package_root` — wrong layer

Adding `.git` to the manifest list changes the semantics of `package_root`. This function finds the *language package root* — the directory containing a build manifest (`Cargo.toml`, `package.json`, `go.mod`, etc.). It's used for:

- Scoping search to the relevant package in a monorepo
- Resolving relative imports
- Blast radius analysis
- MCP scope detection at startup

Adding `.git` makes `package_root` return the git root even for subpackages in a monorepo. For example, in a monorepo with `repo/.git` and `repo/packages/api/package.json`, calling `package_root("repo/packages/api/src/foo.ts")` should return `repo/packages/api/` (the npm package root), not `repo/` (the git root). With `.git` as a fallback, it would still return `repo/packages/api/` (because `package.json` is found first on the way up), but for a project with NO manifest (e.g., a plain scripts directory), it would expand scope to the entire git repo root — which may be enormous.

The problem the contributor is trying to solve (PHP project with `composer.json` not recognized) is better fixed by just adding `composer.json` to the manifest list — which the PR also does and is correct.

If we want a git-root fallback, it should be a separate function (`git_root`) used explicitly where appropriate, not mixed into `package_root`.

## `composer.json` — yes, good addition

This is correct. PHP/Laravel projects use `composer.json` as their manifest. Should be included.

## `would_outline` reformatting — noise

The `read/mod.rs` change is a whitespace-only reformatting that was already fixed in v0.6.3 (commit `08b006b`). The PR was likely branched before that fix. This will cause a merge conflict or is already stale.

---

## Summary

| Change | Verdict |
|--------|---------|
| `follow_links(false)` | **Reject** — breaks pnpm, workspaces, monorepos, macOS `/tmp`. Too aggressive. |
| `same_file_system(true)` | **Accept** — correctly prevents NFS/cross-device traversal. This is the right fix. |
| `.git` as fallback manifest | **Reject** — wrong layer. `package_root` is for language manifests, not git roots. |
| `composer.json` manifest | **Accept** — correct, PHP projects need this. |
| `would_outline` formatting | **Reject** — already fixed on main, diff noise. |
| Test changes | **Reject** — tests should continue to verify symlink behavior works, not assert it's broken. |

### What I'd accept

A PR with just these two lines:

```rust
// src/search/mod.rs — walker()
builder
    .follow_links(true)
    .same_file_system(true)  // Stop at mount point boundaries (NFS, external volumes)
```

```rust
// src/lang/mod.rs — MANIFESTS
"mix.exs",
"composer.json", // PHP/Laravel
```

The existing symlink tests should stay as-is (asserting symlinks ARE followed). The `same_file_system` change is invisible to the tests since they all run within a single `tempdir`.

Happy to merge a revised PR with just those changes.

jahala:
Hi @quanhavn! did you have time to take a look at the review above? Would you prefer I narrow the PR myself with Opus (keep same_file_system(true) + composer.json, drop the other changes)? Just want to make sure that’s okay with you.

jahala:
Closing in favor of #110 which landed the narrower scope we agreed on in the [April 21 review](#issuecomment-2820720013). Thanks again for catching the NFS issue, @quanhavn — your fix is on `main` as of 317aa2a with attribution preserved via `Co-Authored-By`. 🙏

PR #66, opened 2026-04-28
Usability patches

choongng:
Six small patches that improve tilth's behavior under both interactive shell and agent (MCP) use.

1. Respect .gitignore by default; add .tilthignore and `--no-ignore` as escapes.
2. Make `--full` imply expand-all on search queries. It was silently a no-op on non-file-path queries.
3. Detect terminal height via TIOCGWINSZ. `terminal_height()` read only `$LINES` and fell back to 24, so any terminal that doesn't export `$LINES` to children (e.g. VS Code) paged spuriously.
4. Don't tag quoted-code mentions as definitions; recognize markdown headings. The keyword heuristic ran on every non-tree-sitter file, including markdown and structured data, so symbol mentions inside fenced blocks were treated as defs.
5. Ensure CLI output ends with a newline. `emit_output` used `print!`; a few formatters (notably the search-result footer) don't terminate, which left the next prompt mid-line.
6. mcp: thread the `.gitignore` / `.tilthignore` / `TILTH_NO_IGNORE` guidance through `tilth_search` and `tilth_deps` descriptions (follow up to #1).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -689,6 +689,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "terminal_size"
+version = "0.4.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "230a1b821ccbd75b185820a1f1ff7b14d21da1e442e22c0863ea5f08771a8874"
+dependencies = [
+ "rustix",
+ "windows-sys",
+]
+
+[[package]]
 name = "tilth"
 version = "0.7.0"
 dependencies = [
@@ -707,6 +717,7 @@ dependencies = [
  "serde_json",
  "streaming-iterator",
  "tempfile",
+ "terminal_size",
  "toml",
  "tree-sitter",
  "tree-sitter-c",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,6 +64,9 @@ toml = "0.8"
 # MCP protocol (JSON-RPC over stdio)
 # (handled manually — no framework needed)
 
+# Terminal size detection (TIOCGWINSZ ioctl, falls back to env vars)
+terminal_size = "0.4"
+
 
 [dev-dependencies]
 tempfile = "3"
--- a/src/index/symbol.rs
+++ b/src/index/symbol.rs
@@ -72,14 +72,9 @@ impl SymbolIndex {
         // We use WalkBuilder for directory filtering but rayon for parallelism
         // because rayon gives us better work-stealing than ignore's parallel walker
         // for CPU-bound tree-sitter parsing.
-        let files: Vec<PathBuf> = WalkBuilder::new(scope)
-            .follow_links(true)
-            .hidden(false)
-            .git_ignore(false)
-            .git_global(false)
-            .git_exclude(false)
-            .ignore(false)
-            .parents(false)
+        let mut builder = WalkBuilder::new(scope);
+        crate::search::apply_ignore_settings(&mut builder);
+        let files: Vec<PathBuf> = builder
             .filter_entry(|entry| {
                 if entry.file_type().is_some_and(|ft| ft.is_dir()) {
                     if let Some(name) = entry.file_name().to_str() {
--- a/src/lang/outline.rs
+++ b/src/lang/outline.rs
@@ -165,9 +165,32 @@ fn node_to_entry(
             (OutlineKind::Import, text, None)
         }
 
-        // Exports
+        // Exports — `export` is a modifier on a wrapped declaration, not a
+        // peer of `function`/`class`/`const`. Recurse into the inner
+        // declaration so the entry renders with its real kind. Falling back to
+        // `OutlineKind::Export` only when there is no nameable declaration
+        // inside (`export { … }`, `export * from …`, `export default <expr>`).
+        // Without this, `export_statement`'s `name` is the full source span
+        // (already starts with `export `), and the renderer prepends the
+        // `Export` kind_label `"export"` again — producing the doubled-keyword
+        // outline header `export export async function foo(`.
         "export_statement" => {
-            let name = node_text(node, lines);
+            let mut cursor = node.walk();
+            for child in node.children(&mut cursor) {
+                if let Some(mut inner) = node_to_entry(child, lines, lang, depth) {
+                    // Extend the entry's range to cover the `export` keyword
+                    // so the outline byte range still points at the statement.
+                    inner.start_line = start_line;
+                    return Some(inner);
+                }
+            }
+            // No nameable inner declaration; strip leading `export ` so the
+            // rendered name doesn't duplicate the `kind_label`.
+            let raw = node_text(node, lines);
+            let name = raw
+                .strip_prefix("export ")
+                .map(str::to_string)
+                .unwrap_or(raw);
             (OutlineKind::Export, name, None)
         }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -430,13 +430,19 @@ fn emit_output(output: &str, is_tty: bool) {
 }
 
 fn terminal_height() -> usize {
-    // Try LINES env var first (set by some shells)
+    // ioctl(TIOCGWINSZ) on stdout — the real terminal height. Bash maintains
+    // $LINES as an unexported shell variable, so most subprocesses (us included)
+    // never receive it; relying on it alone made tilth assume 24 rows in any
+    // typical interactive shell and page nearly every result. Fall back to
+    // $LINES then to 24 only if the ioctl is unavailable (tests, exotic TTYs).
+    if let Some((_, terminal_size::Height(h))) = terminal_size::terminal_size() {
+        return h as usize;
+    }
     if let Ok(lines) = std::env::var("LINES") {
         if let Ok(h) = lines.parse::<usize>() {
             return h;
         }
     }
-    // Fallback
     24
 }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -28,7 +28,15 @@ struct Cli {
     #[arg(long)]
     budget: Option<u64>,
 
-    /// Force full output (override smart view).
+    /// Force full output (effect depends on query type — see --help).
+    ///
+    /// File path: return the whole file instead of an outline (bypass smart view).
+    ///
+    /// Symbol / text / regex: inline source for every match (equivalent to
+    /// `--expand=<all>`). Explicit `--expand=N` wins. Output stays bounded
+    /// by `--budget`.
+    ///
+    /// Glob: no effect (glob queries already return a flat file list).
     #[arg(long)]
     full: bool,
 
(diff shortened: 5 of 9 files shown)


jahala:
Hey @choongng - Thank you for contributing

Here is the review by Opus 4.7

---- 

Thanks for the patches — several of these address real friction points. But "Usability patches" undersells what's actually here. This is 8 commits, 459 lines, and includes one major behavior change, one ~200-line new feature, and one real bug fix mixed in with smaller polish items. Before I can merge, the architecturally questionable bits need rework and the bundle needs splitting.

---

## Mergeable as-is (small, focused, correct)

### Patch 5 — emit trailing newline in `emit_output`

Real bug. `print!` doesn't terminate; some formatters (search footer) don't either. The guard at the sink is the right place. Ship it.

### Patch 3 — terminal height via TIOCGWINSZ

Real fix. `$LINES` is rarely exported to subprocesses, so `terminal_height()` was effectively hardcoded to 24. Adds `terminal_size = "0.4"` — small, no new transitive deps (rustix and windows-sys already in the lockfile). Ship it.

### Commit 8 — `OutlineKind::Export` doubling fix

Real bug. `export function foo() {}` was rendering as `export export async function foo(...)` because `export_statement.name` was the full source span (already starting with `export `) and the renderer prepended the `Export` kind label again. Recursing into the inner declaration and extending `start_line` to cover the `export` keyword is the correct fix. Ship it.

---

## Needs rework

### Patch 4 — markdown heading defs + skip code-block mentions

Direction is right — the keyword heuristic was running on every non-tree-sitter file, producing false-positive "definitions" for any `function foo` mentioned inside a markdown code block. Soft-definition for ATX headings with `def_weight: 30` (between usages 0 and code defs 60-80) is a sensible tier.

But:

1. **Heading match is `line.contains(query)`** — too loose. A heading `## How to call the function` would be tagged as a definition for query `function`. Should match the heading *text* (after stripping `#`s) for a word boundary or exact identifier, not substring-anywhere.
2. **No tests.** A change to definition-detection semantics across all markdown files needs at least: heading-with-query matches, heading-without-query doesn't, indented headings, code-block mentions don't match.

Otherwise this is good — happy to merge with those two cleaned up.

### Patch 2 — `--full` implies `expand=usize::MAX` for searches

Real UX issue (silent no-op on search queries was annoying). But `usize::MAX` is the wrong fix:

- On a 16k-file repo, `tilth foo --full` would now try to compute the expanded source for every match before truncating to budget. That's an enormous amount of wasted parsing/rendering work.
- The output ends up budget-bounded anyway, so the user doesn't see thousands of expansions — but tilth still pays the cost.

Use a reasonable cap — `expand = 50` or so — when `--full` is set with no explicit `--expand`. That's enough to cover any practical "show me everything that matters" case without the unbounded work for pathological queries.

---

## Should not merge as-is

### Patch 1 — `apply_ignore_settings` (gitignore default flip)

This inverts a deliberate, documented design choice. The current code has this comment:

> // We skip these explicitly instead of relying on .gitignore so that locally-relevant
> // gitignored files (docs/, configs, generated code) are still searchable.

The PR flips the default to respect `.gitignore`, with `.tilthignore` `!path` lines as the escape hatch. The PR description frames this as a usability win. It's not — it's a design call with real consequences:

- `.env.example`, `tsconfig.local.json`, generated TypeScript types in `dist/`, build outputs containing docs, vendored snippets — all become invisible by default
- Users with non-trivial `.gitignore` files (most real projects) lose searchability for legitimately useful local files
- The `.tilthignore` workaround requires every user to learn a new file format and manually re-include things that worked before

If we want to flip this default, that's a discussion to have with benchmarks and explicit reasoning, not a patch in a "usability" PR. I'd revert this to opt-IN: keep current default, add `--ignore` / `TILTH_IGNORE=1` for users who want to honor gitignore.

Patch 6 (MCP tool description updates) goes away if we revert this — it's just propagating the new behavior into docs.

### Patch 7 — enclosing scope annotation for usages (~200 lines)

This is the biggest concern. It re-implements something we already have, with worse architecture.

**The duplication:** `src/search/callers.rs:437` already has `find_enclosing_function` that takes a tree-sitter node and walks up via the AST to find the enclosing definition, qualified with its enclosing type (e.g. `Module.func`, `Class.method`). It handles all 13 supported languages including Elixir's `call`-node weirdness. It uses real AST data.

The PR adds a parallel implementation (`code_enclosing_scope` + `parse_outline_kind_and_name`) that:

1. Calls `get_outline_str()` to get the outline string
2. Parses the outline string back into structure via regex-ish text manipulation
3. Has hardcoded lists of "structural" keywords (`function`, `class`, `struct`, `interface`, `trait`, `enum`, `type`, `const`, `let`, `val`, `var`, `fn`, `def`, `fun`, `object`, `mod`, `module`) and "modifiers" (`export`, `default`, `async`, `public`, `private`, `protected`, `static`, `abstract`, `pub`, `extern`, `unsafe`)
4. Uses string heuristics to extract identifiers from rendered outline lines

This is broken architecture. We have AST data, we render it to a string for display, and now we're parsing that string back to recover the AST data we already had. The hardcoded keyword lists are also incomplete — Elixir's `defmodule`/`def`/`defp`/`defmacro` aren't in the structural list, Swift's `init`/`deinit`/`subscript` aren't there, etc. Every new language we add (and every existing one this list missed) silently produces incorrect or empty scope labels.

The right approach is to either:

- **(a)** Generalize the existing `callers::find_enclosing_function` to take a (path, line) and re-parse the file with tree-sitter, returning the enclosing definition. Then call that from search/mod.rs. One implementation, AST-correct, language-aware.
- **(b)** Store the enclosing scope in `OutlineEntry` directly when we build it, so the renderer can attach it without re-parsing anything.

The markdown variant (`markdown_enclosing_scope`) is a separate concern and is fine — markdown has no AST in tilth, finding the nearest preceding ATX heading via line walk is reasonable.

**Recommendation:** drop this patch from the PR. If you want the feature, open a focused PR that wires it through `find_enclosing_function` (option a), or the outline builder (option b). Either approach is straightforward and produces correct results across all languages.

---

## Process notes

- **Bundle is mislabeled.** "Usability patches" implies small fixes. Patches 1, 2, 7 are each substantial enough to merit their own PR with their own discussion.
- **Zero tests added** for 459 lines of changes — including a default behavior flip (gitignore) and a 200-line new feature (scope annotation). The bug fixes (newline, terminal height, export doubling) are arguably small enough not to need them, but the behavior changes need test coverage before merging.
- **Last two commits have empty author field.** Probably a git config issue — worth fixing, but not a blocker.

---

## Suggested split

1. **PR-A**: Patch 5 (newline) + Patch 3 (terminal height) + Commit 8 (export doubling). Three small, correct fixes. Mergeable immediately.
2. **PR-B**: Patch 4 (markdown headings) with tighter heading-text matching + tests.
3. **PR-C**: Patch 2 (`--full` for searches) with sensible cap instead of `usize::MAX`.
4. **PR-D**: Patch 1 (gitignore default) — separate discussion, opt-in not opt-out.
5. **PR-E**: Patch 7 (scope annotation) — rewrite via `find_enclosing_function` or `OutlineEntry` extension.

Happy to merge PR-A as soon as it's split out. The others I'm happy to review individually with the changes above.

choongng:
Thanks for taking a look. If the separate PRs don't make too much work for you I'll break them up roughly as above.

jahala:
Hey @choongng - see the new Opus review below, thank you

---

Thanks for the v2. Splitting out the scope-annotation feature into #72 was the right call — that one's structurally complete. For this PR, the new export-doubling fix is good and three of the patches are mergeable today, but several items I flagged in the prior review still need real changes.

I'd suggest splitting into two: a small "easy wins" PR with the three clean patches, and the remaining items reworked.

## Mergeable as-is

These three are good and ready to land:

- **Patch 3** — terminal_size TIOCGWINSZ. Correct fix; `$LINES` was rarely propagated to subprocesses, so paging at 24 rows kicked in for nearly every interactive run.
- **Patch 5** — emit_output trailing newline. Correct guard at the sink rather than auditing every formatter.
- **Patch 8** (new) — `OutlineKind::Export` doubling fix. Recursing into `export_statement` children, patching `inner.start_line` to cover the `export` keyword, falling back to `strip_prefix("export ")` for `export { … }` and `export default expr`. One side effect worth knowing: an exported class now collects its methods in the outline because the recursion happens at `depth=0`. That's the desired behavior, just worth a sentence in the commit message.

These could land as a small focused mini-PR today.

## Still need rework

### `--full` cap is `usize::MAX`

```rust
(None, true) => usize::MAX,
```

On a 16k-file repo, `tilth foo --full` now attempts to expand every match before budget truncation kicks in. Output stays bounded by `--budget`, but the parsing/rendering work is real and unbounded. The previous review asked for a concrete cap (~50). Replace `usize::MAX` with `50` or similar.

### `TILTH_NO_IGNORE` propagation via `std::env::set_var`

```rust
if cli.no_ignore {
    std::env::set_var("TILTH_NO_IGNORE", "1");
}
```

`std::env::set_var` has UB semantics in multithreaded contexts (Rust upgraded this to `unsafe` in nightly 1.81; stable kept the safe signature but the documented behavior is unchanged). The MCP server path uses rayon and `ignore`'s parallel walker; both spawn threads. `apply_ignore_settings` reads `TILTH_NO_IGNORE` at every walker construction, so a write racing a read in any worker is technically UB.

In practice the set_var fires at startup before threads spawn, so the race window is narrow — but the code is structured to not rely on that ordering. Cleanest fix: parse the env var (and CLI flag) once into a `OnceLock<bool>` at startup, remove `set_var` entirely, have `apply_ignore_settings` read the parsed bool.

### Patch 4 markdown heading match still uses raw substring

```rust
if !line.contains(query) { continue; }
```

`line` includes the leading `#`s. Query `function` matches `## How to call the function`. Strip the hashes, lowercase both sides, do a word-boundary or identifier-boundary check:

```rust
let heading = trimmed.trim_start_matches('#').trim();
let lower_heading = heading.to_ascii_lowercase();
let lower_query = query.to_ascii_lowercase();
if !word_boundary_match(&lower_heading, &lower_query) { continue; }
```

Also, no tests for `find_defs_markdown_buf` or `is_markdown_heading` — the new code path has zero coverage.

### Patch 1 design questions still open

These are design calls more than code calls — would value an explicit decision either way:

- **Silent data loss when `.gitignore` excludes the project's own `src/`**: monorepos that gitignore generated source trees now return zero search results with no indication. Worth emitting a one-line stderr message when the walker's gitignore filter actually excluded files: `> 247 files excluded by .gitignore (use --no-ignore to include)`. The `ignore` crate exposes filtering events.

- **`.tilthignore` parse failures are silent**: malformed rules drop with no warning. A typo like `! src/generated` (extra space) becomes a no-op. Worth a startup validation pass that logs warnings to stderr. Important because `.tilthignore` is the documented escape hatch.

- **MCP-session limitations**: The MCP server doesn't accept a `--no-ignore` flag, so MCP users must set `TILTH_NO_IGNORE=1` in the environment that launches the server. The MCP tool description mentions this as "a one-off bypass" without explaining how to actually set it. Worth either (a) adding a `tilth_set_ignore` MCP tool, or (b) clarifying the docs.

## Suggested split

**PR-A (small, mergeable now)**: patches 3, 5, 8 — terminal height + trailing newline + export doubling fix.

**PR-B (this PR, rebased)**: patches 1, 2, 4, 6 with the rework above. Once the gitignore design questions are settled, this is a reasonable scope for a second PR.

Happy to merge PR-A whenever it's split out. Reach out if any of the design questions on Patch 1 need more discussion.

jahala:
Heads up — PRs #67–#71 just merged into main, so this PR will need a rebase.

Expected textual conflict on `src/search/symbol.rs` between Patch 4 (markdown heading defs) and #67 (skip-minified-files filter sits nearby). Should be small. Other patches are in different files and shouldn't conflict.

Will follow up on the actual review feedback shortly.

PR #72, opened 2026-04-29
search: annotate usages with enclosing scope via tree-sitter

choongng:
Replaces the outline-string heuristic in `format_grouped_usages` (which fell out as `{` for `class Foo {`, returning the last whitespace-split token) and adds the same annotation to `format_single_match`. Code matches now render `[usage in function bar]` / `[N usages in function Class.method]`; markdown matches render `[usage in §Heading]`.

The code path goes through `callers::enclosing_definition_at(path, line)` — a new public entry point that re-parses the file with tree-sitter (cached on `OutlineCache`) and walks up from the deepest node containing `line` to the nearest definition, qualifying with the enclosing type/module when one wraps it. The walker is factored out of the existing `find_enclosing_function` so callers and the formatter share one AST-correct implementation; `find_enclosing_function` keeps its `(name, range)` signature for the existing call sites in `find_callers_treesitter`.

Why this shape

The natural alternative — parse the rendered outline string back into `(kind, name)` via hardcoded keyword/modifier lists — would carry two problems: (1) every new language grammar requires updating the keyword set or it silently produces wrong scope labels, including existing-language gaps like Elixir's `defmodule`/`def`/`defp` and Swift's `init`/`deinit`/`subscript`; (2) it duplicates information the AST already has. Going through tree-sitter avoids both — language coverage is automatic, kind labels come from a single match on AST node kinds (with one Elixir-specific branch since `def` is a `call` node, not a definition kind).

Cache shape

`OutlineCache` already keys on `(path, mtime)` and stores rendered outline strings. Extended it with a parallel `DashMap` of `Arc<ParsedFile>` (content + tree + lang) under the same key discipline. Same lifecycle, no new threading through call sites; if someday this needs to split into its own type the move is mechanical.

Markdown

Markdown has no AST in tilth, so `markdown_enclosing_scope` walks backwards to the nearest preceding ATX heading and truncates to ~60 chars with an ellipsis. Kept narrow so it can't drift into other markdown semantics.

Tests

Seventeen new tests:

- AST walk: Rust top-level + mod-qualified, TypeScript class methods, Python `class.def`, Elixir `Module.func`, Elixir `defmodule` kind.
- Kind labels: parameterized coverage of `class` / `struct` / `enum` / `interface` / `trait` so a regression on the kind_label match arm surfaces immediately.
- Pinned behavior: `impl Foo { fn bar() }` produces `"bar"`, not `"Foo.bar"` — tree-sitter-rust's `impl_item` exposes its type via a `type` field that `extract_definition_name` doesn't probe. Pre-existing in `find_enclosing_function`; the test makes any future fix an intentional, visible change.
- Edges: line=0, non-code, before-first-heading, markdown heading truncation.
- Integration: `format_single_match` round-trips an actual Match through the formatter and asserts the output contains `[usage in function Foo.bar]`.

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -5,24 +5,38 @@ use std::time::SystemTime;
 use dashmap::mapref::entry::Entry;
 use dashmap::DashMap;
 
+use crate::types::Lang;
+
 /// Cached outline entry.
 struct CacheEntry {
     outline: Arc<str>,
 }
 
+/// File contents and its tree-sitter parse, cached together so AST consumers
+/// don't re-parse on every call. `content` is `Arc<String>` so callers can
+/// hold the bytes for `Node::utf8_text` without copying.
+pub struct ParsedFile {
+    pub content: Arc<String>,
+    pub tree: tree_sitter::Tree,
+    pub lang: Lang,
+}
+
 /// Outline cache keyed by (canonical path, mtime). If the file changes,
 /// mtime changes and the old entry is never hit again.
 ///
-/// Value is `Arc<str>` — inline string data in the Arc allocation,
-/// one less indirection than `Arc<String>`.
+/// Stores two derived analyses: rendered outline strings (used by search
+/// formatting) and parsed tree-sitter trees (used by AST scope queries).
+/// Both share the same key + invalidation; nothing else is shared.
 pub struct OutlineCache {
     entries: DashMap<(PathBuf, SystemTime), CacheEntry>,
+    parsed: DashMap<(PathBuf, SystemTime), Arc<ParsedFile>>,
 }
 
 impl Default for OutlineCache {
     fn default() -> Self {
         Self {
             entries: DashMap::new(),
+            parsed: DashMap::new(),
         }
     }
 }
@@ -49,6 +63,38 @@ impl OutlineCache {
                     outline: Arc::clone(&outline),
                 });
                 outline
+            }
+        }
+    }
+
+    /// Parse a code file with tree-sitter and cache the result. Returns
+    /// `None` for non-code files, files larger than the 500 KB cap, or parse
+    /// failures.
+    #[must_use]
+    pub fn get_or_parse(&self, path: &Path) -> Option<Arc<ParsedFile>> {
+        let meta = std::fs::metadata(path).ok()?;
+        let mtime = meta.modified().ok()?;
+        if meta.len() > 500_000 {
+            return None;
+        }
+        match self.parsed.entry((path.to_path_buf(), mtime)) {
+            Entry::Occupied(e) => Some(Arc::clone(e.get())),
+            Entry::Vacant(e) => {
+                let crate::types::FileType::Code(lang) = crate::lang::detect_file_type(path) else {
+                    return None;
+                };
+                let ts_lang = crate::lang::outline::outline_language(lang)?;
+                let content = std::fs::read_to_string(path).ok()?;
+                let mut parser = tree_sitter::Parser::new();
+                parser.set_language(&ts_lang).ok()?;
+                let tree = parser.parse(&content, None)?;
+                let parsed = Arc::new(ParsedFile {
+                    content: Arc::new(content),
+                    tree,
+                    lang,
+                });
+                e.insert(Arc::clone(&parsed));
+                Some(parsed)
             }
         }
     }
--- a/src/search/callers.rs
+++ b/src/search/callers.rs
@@ -7,7 +7,10 @@ use std::sync::{Arc, Mutex};
 
 use streaming_iterator::StreamingIterator;
 
-use crate::lang::treesitter::{extract_definition_name, DEFINITION_KINDS};
+use crate::lang::treesitter::{
+    extract_definition_name, extract_elixir_definition_name, is_elixir_definition,
+    node_text_simple, DEFINITION_KINDS,
+};
 
 use crate::cache::OutlineCache;
 use crate::error::TilthError;
@@ -442,8 +445,6 @@ fn find_callers_treesitter_batch(
     callers
 }
 
-/// Walk up the AST from a node to find the enclosing function definition.
-/// Returns (`function_name`, `line_range`).
 /// Type-like node kinds that can enclose a function definition.
 const TYPE_KINDS: &[&str] = &[
     "class_declaration",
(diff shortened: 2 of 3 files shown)


jahala:
Hi @choongng - please see the Opus review below

---

Thanks — this is a model architectural response. Tree-sitter direct, factored shared walker, point-query, 17 tests including a pinned-behavior test for the `impl Foo { fn bar() }` gap. I read through the code carefully and ran concrete inputs through the markdown path. Findings below.

The core architecture is correct. `tree_sitter::Tree: Sync` holds in 0.25 (confirmed by compile), the `Arc<ParsedFile>` bundling is load-bearing (`Node::utf8_text` requires the same source bytes that were parsed), the cache TOCTOU is the same shape as the existing outline cache, and `walk_to_enclosing_definition` preserves the callers-path behavior.

## Must fix before merge

### `markdown_enclosing_scope` has three CommonMark gaps

I ran these inputs through a standalone copy of the function and got the wrong output for all three. They're not theoretical — concrete bugs:

```text
"####### Fake Heading 7"   →  Some("§# Fake Heading 7")
   spec: 7+ leading `#`s is NOT a heading (CommonMark §4.6.1)

"##NoSpace"                →  Some("§NoSpace")
   spec: `#`s must be followed by space or EOL

"## Real\nfoo\n```md\n## Inside Fence\n...\n```\n..."  match_line=5
                           →  Some("§Inside Fence")
   spec: ATX heading inside a fenced code block is literal content
```

The current loop caps the count at 6 but doesn't reject 7+, and there's no whitespace check after the hashes. Suggested rewrite:

```rust
let hashes = trimmed.bytes().take_while(|&b| b == b'#').count();
if !(1..=6).contains(&hashes) { continue; }
match trimmed.as_bytes().get(hashes) {
    None | Some(b' ' | b'\t') => {}
    _ => continue,
}
let text = trimmed[hashes..].trim_start_matches([' ', '\t']).trim_end();
```

Plus a fence-state pass for the third case — either a forward scan to build a "lines-inside-fences" bitset, or use the same fence tracking that `resolve_heading` already implements (so the duplication is on a real shared helper rather than scattered).

### Reconcile heading detection with PR #66

PR #66 has its own `is_markdown_heading` in `src/search/symbol.rs` that does the same parse with slightly different shape. If both PRs land separately, day one has two ATX-heading parsers in the codebase. They should call a shared helper before either lands. Either consolidate in this PR or as a precursor commit before the second one merges.

### Cache test comment doesn't match what's tested

`enclosing_at_caches_parse_across_calls` claims "the second sees stale data because the mtime didn't change" but the test doesn't mutate the file between the two calls. As written, it only proves consistency. To prove cache hits via stale-data observation, you'd need a `filetime` dev-dep to mutate content while keeping mtime constant. Either rewrite the comment or add the actual stale-data test.

## Worth documenting

### Behavior change for anonymous definitions

The old `find_enclosing_function` had:
```rust
let name = extract_definition_name(n, lines).unwrap_or_else(|| "<anonymous>".to_string());
```
which terminated the walk at any DEFINITION_KIND node, returning `"<anonymous>"` when no name could be extracted. The new `walk_to_enclosing_definition` walks PAST that node looking for an outer named definition.

This is probably an improvement (more useful scope info for callers reports), but it's a silent behavior change. Worth adding a note to the PR description and a test pinning the new behavior — e.g., a JS anonymous function nested inside a named function should now report the named function as the enclosing scope, not `"<anonymous>"`.

### `Tree: Sync` version dependency

Add a one-line comment in `src/cache.rs::ParsedFile`:
```rust
// `tree_sitter::Tree` implements `Send + Sync` since 0.25 — required for
// `Arc<ParsedFile>` storage in `DashMap`. Downgrading below 0.25 breaks
// this at compile time.
```

A future cargo update that downgraded would fail to compile, but the failure mode would be obscure without the comment pointing at the version requirement.

## Non-blocking

- **Memory growth**: `parsed: DashMap` is unbounded. Heavy MCP sessions on large repos will accumulate parsed trees indefinitely. Same shape as the existing outline cache, but per-entry cost is higher (~100KB-1MB vs. a few KB). Worth a tracking issue, not a blocker.

- **Location of `enclosing_definition_at`**: Lives in `callers.rs` but is conceptually a scope-annotation operation, not a callers operation. A future `src/search/scope.rs` would be a cleaner home. Doesn't block.

- **Inline definition test**: Worth adding `fn foo() { bar(); }` on a single line — current behavior (returns `foo` as scope) is correct, but a test pins it.

## What's right

- Architecture matches what the prior review asked for — tree-sitter direct, factored walker, no string-parsing back from rendered outline.
- `Arc<ParsedFile>` bundling is correct — content and tree must travel together for `utf8_text` consistency.
- Test coverage is wide: parameterized kind labels, language spread (Rust, TypeScript, Python, Elixir), integration round-trip through `format_single_match`, pinned-behavior test for the `impl Foo` gap.
- `kind_label` is exhaustive enumeration, not string heuristics — good shape for adding new languages.
- Existing 256 tests still pass on test-merge with main, clippy clean.

Looking forward to merging this once the markdown gaps are tightened up. Happy to extract a precursor "shared ATX-heading helper" commit if it makes the consolidation with #66 easier.

jahala:
Heads up — PRs #67–#71 just merged into main, so this PR will need a rebase.

The notable conflict will be in `src/search/callers.rs`: #69 changed `find_callers` from `Result<Vec<CallerMatch>, TilthError>` to `Result<(Vec<CallerMatch>, bool), TilthError>` (added a `target_seen` flag). Existing call sites need to destructure the tuple. Should be a small mechanical update.

Otherwise no semantic conflicts — the new mmap-scan-on-oversize-files behavior in `find_callers` doesn't touch `enclosing_definition_at`, which has its own `OutlineCache::get_or_parse` path.

Will follow up on the actual review feedback shortly.

choongng:
I think for the time being I will continue development on the branch linked below, but feel free to pull anything that seems useful.
https://github.com/choongng/tilth/commits/dev/

jahala:
Resolved via #77 — pulled the enclosing-scope annotation from this PR (`7cad3f1`) along with the 14-commit markdown rewrite from your dev/ branch (16 cherry-picks total + 4 follow-up commits from us).

The 3 markdown_enclosing_scope bugs flagged in earlier review are now structurally impossible because `c592109` migrates that function (and its sibling `markdown_defs`) to the tree-sitter-md AST. To make sure they can't silently regress through a future grammar swap, our follow-up `5d52013` adds 3 pinned regression tests at the `markdown_enclosing_scope` layer (7-hash, no-space, in-fence cases).

We've also pulled additional dev/ work via:
- #75 — 7 standalone fixes (export const detection, path normalization, SymbolIndex removal, etc.)
- #76 — 4 refactor commits (drop unused params, collapse caller path, overview WalkBuilder, manifest serde)
- #78 — multi-glob `tilth_files`

Plus PR-D (#77) includes `e3c4907` (the bloom_walk/callee_query/scope extraction).

**Held items** documented in PR-D's body — `7e4d963` (`.gitignore` default) needs a `set_var` UB fix before adoption, and `fd3de77` (SERVER_INSTRUCTIONS rewrite) needs haiku benchmarks.

Author preserved on every cherry-pick (`git log --pretty='%an'` confirms). Thanks for sticking with this — the dev/ branch was a goldmine.

Closing in favour of #77.

PR #73, opened 2026-05-01
fix(install): add opencode-native MCP config format

jmarcelomb:
Opencode uses a different MCP server entry format than the standard command/args split. It expects `"type": "local"` with `"command"` as a flat array.

Changes:
- Add `ConfigFormat::JsonLocal` variant for opencode's `{type, command[]}` shape
- `ConfigFormat::Json` remains the standard `{command, args}` shape
- Fixed opencode servers key: `"mcpServers"` → `"mcp"`
- Fixed opencode config path: `~/.opencode.json` → `~/.config/opencode/opencode.json`
- Added `opencode_resolve_host` and `opencode_entry_uses_local_shape` tests

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -53,7 +53,7 @@ const SUPPORTED_HOSTS: &[&str] = &[
     "pi",
 ];
 
-/// The tilth server entry as JSON. Format depends on the host's ConfigFormat variant.
+/// The tilth server entry as JSON. Format depends on the host's [`ConfigFormat`] variant.
 fn tilth_server_entry(edit: bool, format: &ConfigFormat) -> Value {
     let (command, args) = tilth_command_and_args(edit);
     match format {
@@ -69,10 +69,7 @@ fn tilth_server_entry(edit: bool, format: &ConfigFormat) -> Value {
                 "command": command_arr
             })
         }
-        ConfigFormat::Toml => json!({
-            "command": command,
-            "args": args
-        }),
+        ConfigFormat::Toml => unreachable!("tilth_server_entry called for TOML host"),
     }
 }
 
@@ -87,7 +84,7 @@ pub fn run(host: &str, edit: bool) -> Result<(), String> {
 
     match host_info.format {
         ConfigFormat::Json { .. } | ConfigFormat::JsonLocal { .. } => {
-            write_json_config(&host_info, edit)?
+            write_json_config(&host_info, edit)?;
         }
         ConfigFormat::Toml => write_toml_config(&host_info, edit)?,
     }
@@ -105,8 +102,7 @@ pub fn run(host: &str, edit: bool) -> Result<(), String> {
 
 fn write_json_config(host_info: &HostInfo, edit: bool) -> Result<(), String> {
     let servers_key = match host_info.format {
-        ConfigFormat::Json { servers_key } => servers_key,
-        ConfigFormat::JsonLocal { servers_key } => servers_key,
+        ConfigFormat::Json { servers_key } | ConfigFormat::JsonLocal { servers_key } => servers_key,
         ConfigFormat::Toml => unreachable!("write_json_config called for TOML host"),
     };
 
@@ -270,9 +266,7 @@ fn resolve_host(host: &str) -> Result<HostInfo, String> {
         // OpenCode user scope: ~/.config/opencode/opencode.json → mcp (local entry shape)
         "opencode" => Ok(HostInfo {
             path: home.join(".config/opencode/opencode.json"),
-            format: ConfigFormat::JsonLocal {
-                servers_key: "mcp",
-            },
+            format: ConfigFormat::JsonLocal { servers_key: "mcp" },
             note: Some("User scope — available in all projects."),
         }),
 
@@ -293,6 +287,7 @@ fn resolve_host(host: &str) -> Result<HostInfo, String> {
         }),
 
         // Amp user scope: ~/.config/amp/settings.json → amp.mcpServers
+        // Verified from official docs: https://ampcode.com/manual
         "amp" => Ok(HostInfo {
             path: home.join(".config/amp/settings.json"),
             format: ConfigFormat::Json {
@@ -302,6 +297,7 @@ fn resolve_host(host: &str) -> Result<HostInfo, String> {
         }),
 
         // Google Antigravity user scope: ~/.gemini/antigravity/mcp_config.json → mcpServers
+        // Verified from official docs: https://antigravity.google/docs/mcp
         "antigravity" => Ok(HostInfo {
             path: home.join(".gemini/antigravity/mcp_config.json"),
             format: ConfigFormat::Json {
@@ -311,6 +307,7 @@ fn resolve_host(host: &str) -> Result<HostInfo, String> {
         }),
 
         // Factory Droid user scope: ~/.factory/mcp.json → mcpServers
+        // Verified from official docs: https://docs.factory.ai/cli/configuration/mcp
         "droid" => Ok(HostInfo {
             path: home.join(".factory/mcp.json"),
             format: ConfigFormat::Json {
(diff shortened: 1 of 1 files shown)


jahala:
Hey @jmarcelomb  - thank you so much for contributing to tilth! 

As I'm not much of a dev, I use Opus for coding etc. Please have a look at the review from Opus 4.7: 

---

Verified the three claims against opencode.ai docs — path `~/.config/opencode/opencode.json`, key `mcp`, and local entry shape `{type: "local", command: [...]}` all match. The fix itself is correct.

Two things before merge:

## 1. Add an opencode test

The existing convention in `install.rs` is one `<host>_resolve_host()` test per host (see `amp_resolve_host`, `droid_resolve_host`, `antigravity_resolve_host`). The PR changes opencode's path, servers key, AND entry shape — three claims that should be pinned by a test. Without one, a future opencode grammar tweak silently breaks the integration again.

```rust
#[test]
fn opencode_resolve_host() {
    let info = resolve_host("opencode").expect("opencode should resolve");
    assert!(info.path.ends_with(".config/opencode/opencode.json"));
    match info.format {
        ConfigFormat::JsonLocal { servers_key } => assert_eq!(servers_key, "mcp"),
        _ => panic!("opencode should use JsonLocal format"),
    }
}

#[test]
fn opencode_entry_uses_local_shape() {
    let entry = tilth_server_entry(false, &ConfigFormat::JsonLocal { servers_key: "mcp" });
    assert_eq!(entry["type"], "local");
    assert!(entry["command"].is_array());
    assert!(entry.get("args").is_none());
}
```

## 2. Fold entry shape into `ConfigFormat`

`EntryStyle` + the new `HostInfo.entry_style` field + 17 lines of `entry_style: EntryStyle::Standard,` across the resolve arms is more surface area than the change needs. The entry shape is coupled to the host's config format — they don't vary independently — so put them together:

```rust
enum ConfigFormat {
    Json { servers_key: &'static str },        // command + args
    JsonLocal { servers_key: &'static str },   // type + command[] (opencode)
    Toml,
}
```

That removes the `EntryStyle` enum, the new field on `HostInfo`, and 17 boilerplate lines from the resolve arms. `tilth_server_entry` switches on `ConfigFormat` directly. Same behavior, less code.

## Style notes

- `&EntryStyle` is borrowed but `EntryStyle` is `Copy` (1-byte enum). Idiomatic Rust passes `Copy` types by value. If you keep `EntryStyle`, change the parameter to `entry_style: EntryStyle`. (Moot if you fold into `ConfigFormat`.)
- The new opencode comment is multi-line; every other arm in this file uses a single-line `// Host: ~/path → key`. Match the existing style for consistency.

## What's right

- Doc comment at line 14 correctly updated to the new path.
- Blast radius is contained — only the opencode arm changes behavior; other hosts get `EntryStyle::Standard` boilerplate but no semantic change.
- Citation in the PR description is verifiable (matches opencode.ai/docs/mcp-servers/).

Otherwise good. Ping when those are addressed.

jmarcelomb:
Addressed both review points:

1. **Folded into `ConfigFormat`** — `EntryStyle` enum and `HostInfo.entry_style` field removed. Added `ConfigFormat::JsonLocal { servers_key }` variant instead. Only opencode uses it; all other hosts stay on `Json`. Net -17 lines.

2. **Added opencode tests** — `opencode_resolve_host` (path + format + servers_key), `opencode_entry_uses_local_shape`, `opencode_entry_with_edit`, and `opencode_upserts_under_mcp_key`.

Also fixed the `&EntryStyle` borrow — `ConfigFormat` is now matched directly by `tilth_server_entry`, no more reference-to-Copy issue.

jahala:
Opus comments: 

---

Refactor is the right shape — `ConfigFormat::JsonLocal` folds the entry shape in cleanly, the `EntryStyle` enum is gone, and the four new tests pin the contract. All 30 install tests pass locally.

But CI will fail on this branch. Three blockers:

## 1. `cargo fmt --check` fails (10 diffs)

```
$ cargo fmt --check 2>&1 | grep -c "^Diff"
10
```

Examples — these were correctly formatted before, the new code regressed them:

```diff
         "opencode" => Ok(HostInfo {
             path: home.join(".config/opencode/opencode.json"),
-            format: ConfigFormat::JsonLocal {
-                servers_key: "mcp",
-            },
+            format: ConfigFormat::JsonLocal { servers_key: "mcp" },
```

```diff
-            ConfigFormat::JsonLocal { .. } => panic!("antigravity should use Json format, not JsonLocal"),
+            ConfigFormat::JsonLocal { .. } => {
+                panic!("antigravity should use Json format, not JsonLocal")
+            }
```

Run `cargo fmt` once and re-commit.

## 2. Missing trailing newline on `src/install.rs`

```
+}
\ No newline at end of file
```

`cargo fmt` will fix this too.

## 3. `cargo clippy -- -D warnings` fails (4 errors)

```
error: item in documentation is missing backticks
  --> src/install.rs:56
   | /// The tilth server entry as JSON. Format depends on the host's ConfigFormat variant.
                                                                        ^^^^^^^^^^^^
```
→ Wrap in backticks: `` `ConfigFormat` ``.

```
error: these match arms have identical bodies
  --> src/install.rs:60
```
The `Json` and `Toml` arms in `tilth_server_entry` produce identical output. But `tilth_server_entry` is only called from `write_json_config` — the `Toml` arm is dead code. Match the pattern already used in `write_json_config:110`:

```rust
ConfigFormat::Toml => unreachable!("tilth_server_entry called for TOML host"),
```

```
error: consider adding a `;` to the last statement
  --> src/install.rs:90
```
→ Add `;` to `write_json_config(&host_info, edit)?;`.

```
error: these match arms have identical bodies
  --> src/install.rs:108
   | ConfigFormat::Json { servers_key } => servers_key,
   | ConfigFormat::JsonLocal { servers_key } => servers_key,
```
→ Merge with `|`:

```rust
let servers_key = match host_info.format {
    ConfigFormat::Json { servers_key } | ConfigFormat::JsonLocal { servers_key } => servers_key,
    ConfigFormat::Toml => unreachable!("write_json_config called for TOML host"),
};
```

## Out-of-scope cleanup to flag

The diff strips ~13 `// Verified from official docs: <url>` citation comments from other host arms (amp, antigravity, droid, zed, copilot-cli, augment, kiro, kilo-code, cline, roo-code, trae, qwen-code, crush, pi). Those weren't comment-style noise — they were the audit trail showing which upstream doc each format claim came from. When opencode silently breaks again, those citations are how we re-verify the fix. Please restore them — keep the opencode arm single-line as fixed, but leave the others alone.

## What's right

- `ConfigFormat::JsonLocal { servers_key }` is the right factoring — same shape as `Json`, opencode-only entry shape branch in `tilth_server_entry`.
- Tests are exhaustive: `opencode_resolve_host`, `opencode_entry_uses_local_shape`, `opencode_entry_with_edit`, `standard_entry_format`, `opencode_upserts_under_mcp_key`. All pass.
- Adding `JsonLocal { .. } => panic!(...)` arms to existing host tests keeps the variant exhaustive so future regressions surface.
- `tilth_server_entry(edit, &ConfigFormat)` — passing the enum by reference is correct since `ConfigFormat` is not `Copy`.

Push the fmt + clippy fixes and restore the citation comments and this should be good to merge.

jmarcelomb:
All review feedback addressed in 260d84a:

- **cargo fmt** — ran and committed
- **cargo clippy** — 4 fixes:
  - Doc comment: backtick-wrapped `ConfigFormat`
  - `tilth_server_entry` Toml arm → `unreachable!`
  - Semicolon after `write_json_config` call
  - Merged `Json | JsonLocal` match arms
- **Restored citation comments** — all 14 host doc URLs that were stripped in the refactor

jahala (approved):
Re-reviewed after rebase. All 4 commits cherry-picked cleanly onto current main (which now has the Services struct + SymbolIndex removal). Verified locally:

- `cargo build` clean
- `cargo test --lib install::` — 30 install tests pass
- `cargo clippy -- -D warnings` clean
- `cargo fmt --check` clean

Author preserved on all 4 commits. Approved.

Heads-up: I'll do a tiny follow-up PR to fix the README install table, which still says `~/.opencode.json` — caught while reviewing. Not a blocker for this merge.

PR #63, opened 2026-04-20
refactor(mcp): consolidate pipeline deps into Services struct

paulnsorensen:
## Summary

Replaces the 5 individual `Arc<...>` / `bool` parameters threaded through `run` → `handle_request` → `handle_tool_call` → `dispatch_tool` with one `Clone`-able `Services` handle (`cache`, `session`, `index`, `bloom`, `edit_mode`). Inner `tool_*` helpers keep their existing signatures — `dispatch_tool` is the single seam where the struct gets unpacked.

No behaviour change. Reduces parameter lists from 6–7 items to 2 and gives future pipeline dependencies a single place to land.

One commit, one file, +46 / −53 lines.

## Test plan

- [x] `cargo build` — clean
- [x] `cargo test` — 239 tests pass, no regressions
- [x] `cargo fmt --check` — clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #82, opened 2026-05-09
refactor(mcp): extract per-request timeout into its own module

paulnsorensen:
## Summary

Part 2 of the #61 split (after #63's `Services` refactor). Extracts the inline `mpsc::recv_timeout` machinery in `handle_tool_call` into a dedicated `src/timeout.rs` backed by `crossbeam-channel`'s `select! { default(timeout) => ... }` — the sync-world equivalent of Java's `Future.get(timeout, unit)`.

`ThreadTracker` is now owned by `Services` rather than a static global, so unit tests instantiate their own and don't serialise on shared state.

No behaviour change for the happy path: same 90s default, same `TILTH_TIMEOUT` env override, same client-visible "tool timed out" / "tool panicked" messages. New: an upfront `tracker.is_at_cap()` check in `handle_tool_call` refuses new work cleanly under sustained pressure instead of piling on more abandoned threads.

## What I changed vs. the original #61 draft

The two timeout-module concerns from your #61 review are addressed here:

- **#6 (Ordering on `is_at_cap`)** — `ThreadTracker::is_at_cap` now uses `Ordering::Acquire`. It pairs with the `AcqRel` RMWs in `record_timeout` / `record_finish_after_timeout`, so a load that observes the new count also observes any state the incrementing thread published before it.
- **#7 (increment-before-CAS false rejection)** — the deadline arm now CAS-claims the timeout *before* touching the tracker (was: increment, then CAS, then conditionally roll back). The worker, if it lost the CAS, spins on a new `timeout_acked` `AtomicBool` before decrementing — so the tracker can never go negative and a concurrent `is_at_cap()` can never observe an inflated count that's about to roll back. Closes the false-rejection window near the hard cap.

I left #5 (`TILTH_BATCH_TIMEOUT` removal) for the batch-edit / read-parallelisation extraction since that's where it lives.

## Review round 2 fixes (applied in `49370e1`)

- RMWs on the abandoned-thread counter switched from `Release` to `AcqRel` — canonical pessimistic ordering for an RMW that's read from another thread. Doc-comment on `is_at_cap` updated to match.
- `wait_for_timeout_ack` spin loop now calls `std::thread::yield_now()` after `spin_loop()` so a worker scheduled before the main thread on a single-CPU container surrenders the rest of its quantum instead of burning ~10ms before the ack becomes visible.
- `ABANDONED_THREAD_WARN`'s deadline check uses `==` (not `>=`) deliberately — warn once at the threshold, hard-cap stops accepting work past it. The constant's doc-comment now calls this out so a future maintainer doesn't "fix" it back to `>=`.
- `SpawnFailure` is `#[non_exhaustive]` so a future failure mode (e.g. OS-level thread spawn failure) can be added without churning every call site.
- `abandoned_counter_roundtrips_through_cas` deadline relaxed 2s → 5s to reduce flake risk on contended runners.

## Files

- `src/timeout.rs` (new, +316) — `ThreadTracker`, `ThreadCoord`, `spawn_with_timeout`, `request_timeout`, plus 5 unit tests.
- `src/mcp.rs` — drops the inline `mpsc` plumbing; `handle_tool_call` shrinks to a cap-check + `run_tool_with_timeout` + `build_tool_response` split. `Services` gains a private `tracker: Arc<ThreadTracker>` field with a `pub(crate) fn tracker()` accessor matching the encapsulation pattern from #63.
- `src/lib.rs` — registers the new `pub(crate) mod timeout`.
- `Cargo.toml` / `Cargo.lock` — adds `crossbeam-channel = "0.5"`.

No version bump.

## Test plan

- [x] `cargo build` — clean
- [x] `cargo test` — 349 pass (5 in `timeout::tests`: CAS roundtrip, fast path, panic surfacing, saturated tracker, env parsing)

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -156,6 +156,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b05b61dc5112cbb17e4b6cd61790d9845d13888356391624cbe7e41efeac1e75"
 
 [[package]]
+name = "crossbeam-channel"
+version = "0.5.15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "82b8f8f868b36967f9606790d1903570de9ceaf870a7bf9fbbd3016d636a2cb2"
+dependencies = [
+ "crossbeam-utils",
+]
+
+[[package]]
 name = "crossbeam-deque"
 version = "0.8.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -739,6 +748,7 @@ version = "0.8.1"
 dependencies = [
  "clap",
  "clap_complete",
+ "crossbeam-channel",
  "dashmap",
  "globset",
  "grep-regex",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,6 +54,11 @@ memmap2 = "0.9"
 # Parallelism
 rayon = "1"
 
+# Channel with select! for the timeout wrapper (sibling of crossbeam-deque/-utils
+# already pulled in transitively by rayon; gives the closer-to-Java-Future
+# ergonomics that std::sync::mpsc::recv_timeout cannot express on multiple arms).
+crossbeam-channel = "0.5"
+
 # Concurrent cache
 dashmap = "6"
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -29,6 +29,7 @@ pub mod overview;
 pub(crate) mod read;
 pub(crate) mod search;
 pub(crate) mod session;
+pub(crate) mod timeout;
 pub(crate) mod types;
 
 use std::path::Path;
--- a/src/mcp.rs
+++ b/src/mcp.rs
@@ -1,10 +1,7 @@
 use std::fmt::Write as _;
 use std::io::{self, BufRead, Write};
 use std::path::{Path, PathBuf};
-use std::sync::atomic::{AtomicUsize, Ordering};
-use std::sync::mpsc;
 use std::sync::Arc;
-use std::time::Duration;
 
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
@@ -12,6 +9,7 @@ use serde_json::Value;
 use crate::cache::OutlineCache;
 use crate::index::bloom::BloomFilterCache;
 use crate::session::Session;
+use crate::timeout::{self, spawn_with_timeout, SpawnFailure, ThreadTracker};
 
 /// Shared dependencies passed through the request → dispatch pipeline.
 #[derive(Clone)]
@@ -19,6 +17,7 @@ pub(crate) struct Services {
     cache: Arc<OutlineCache>,
     session: Arc<Session>,
     bloom: Arc<BloomFilterCache>,
+    tracker: Arc<ThreadTracker>,
     edit_mode: bool,
 }
 
@@ -28,6 +27,7 @@ impl Services {
             cache: Arc::new(OutlineCache::new()),
             session: Arc::new(Session::new()),
             bloom: Arc::new(BloomFilterCache::new()),
+            tracker: Arc::new(ThreadTracker::new()),
             edit_mode,
         }
     }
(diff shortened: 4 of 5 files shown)


jahala:
Reviewed by Opus 4.7:

The split is clean — `ThreadCoord` + the CAS state machine is genuinely easier to reason about than the original inline `mpsc::recv_timeout` plumbing, and the `claim_timeout` / `wait_for_timeout_ack` choreography correctly closes the false-rejection window from your original review item. New tests cover the right spots.

Three things I want to discuss before stamping. None are clear blockers — more questions where your reasoning probably differs from mine and I'd rather check than assume.

### 1. Memory ordering on the abandoned counter — is `Release` deliberate, or would `AcqRel` be safer?

```rust
// timeout.rs
fn record_timeout(&self) -> usize {
    self.count.fetch_add(1, Ordering::Release) + 1
}
fn record_finish_after_timeout(&self) {
    self.count.fetch_sub(1, Ordering::Release);
}
// is_at_cap reads with Ordering::Acquire
```

For a pure counter that doesn't carry adjacent state, `Release`/`Acquire` works (and `Relaxed`/`Relaxed` would too — atomic increments are atomic regardless of ordering). My only nit is that `AcqRel` is the canonical pessimistic choice for an RMW that's read elsewhere, and on weakly-ordered architectures it documents intent better. If you specifically chose `Release` because there's no other data being published through the counter, totally fine — just call that out in a comment so the next reader doesn't second-guess.

### 2. `wait_for_timeout_ack` spinning without `yield_now()` — intentional, or worth a one-liner?

```rust
fn wait_for_timeout_ack(&self) {
    while !self.timeout_acked.load(Ordering::Acquire) {
        std::hint::spin_loop();
    }
}
```

Your doc-comment is right that this terminates in microseconds under healthy scheduling — `spin_loop()` is a CPU hint, not a no-yield directive, and the OS preempts at quantum boundary regardless. So this isn't a livelock. The thing I'd consider: on a single-CPU container with the worker scheduled before the main thread, the worker burns its full quantum (~10ms) before the main thread gets a turn to set the flag. A `std::thread::yield_now()` at the bottom of the loop body skips that wait at trivial cost on healthy systems. Up to you — preserves your fast-path latency either way.

### 3. Warning threshold change `>=` → `==` — intentional dedup, or accidental regression?

Old: `if abandoned >= ABANDONED_THREAD_WARN { eprintln!(...); }` — fires on every increment past 3, so logs flood as threads pile.
New: `if n == ABANDONED_THREAD_WARN { eprintln!(...); }` — fires exactly once at 3, then silence.

Both behaviors are defensible:
- `==`: one warning at threshold, operator investigates, no log noise. Modern best practice for log-spam-prone events.
- `>=`: escalating signal as pressure increases, operator can see severity from log volume alone.

Just want to make sure you intended to switch policy here, since the doc-comment ("warn on sustained pressure") could read either way. If `==` was intentional, consider amending the doc-comment to call it out as a once-and-done warning so the next maintainer doesn't "fix" it back to `>=`.

### Nits (non-blocking)

- `SpawnFailure` could use `#[non_exhaustive]` for forward-compat — a third failure mode (e.g., OS-level thread spawn failure) could be added later without breaking external consumers.
- The 2-second deadline in `abandoned_counter_roundtrips_through_cas` will likely be fine on local + CI, but is the kind of timing that flakes on a contended runner. 5s or a `Condvar`-based wait would be more robust.

The shape of this is right. The CAS choreography is a real improvement over the original inline plumbing. Once you've thought through the three above (or convinced me the originals are deliberate), this is good to merge.

paulnsorensen:
Round 2 pushed in `49370e1` (PR body has the per-item summary under "Review round 2 fixes"):

1. **Memory ordering** — switched to `AcqRel` on both RMWs. You're right that `Release`/`Acquire` works for a pure counter, but `AcqRel` is the canonical pessimistic choice you flagged and it documents intent better than a "why did we pick Release here?" comment would.
2. **`yield_now()`** — added after `spin_loop()`. Single-CPU container case you described is the right reason; on healthy systems the loop body still rarely executes anyway.
3. **`==` warn-once** — was intentional (avoid log spam). Amended the constant's doc-comment to call it out as deliberate so the next maintainer doesn't switch it back to `>=`.
4. **`#[non_exhaustive]` on `SpawnFailure`** — added. It's `pub(crate)` so the attribute is technically a no-op for external consumers we don't have, but the signal-of-intent is worth keeping.
5. **2s → 5s deadline** — bumped. Skipped the `Condvar` route since the polled-counter approach reads more directly.

PR rebased onto current `main` while I was in there.

PR #84, opened 2026-05-09
feat(mcp): batch tilth_edit — accept files: [{path, edits}]

paulnsorensen:
## Summary

Part 3 of the #61 split — batch \`tilth_edit\`. Independent of #82 (timeout) and #83 (diff overlay rayon); touches \`src/edit.rs\` (the new \`apply_batch\` + \`FileEditTask\`), \`src/mcp.rs\` (new \`tool_edit\` + parsing helpers + schema), and \`AGENTS.md\`.

## What changes for callers

- \`tilth_edit\` now takes \`{\"files\": [{\"path\", \"edits\"}, ...]}\` instead of \`{\"path\", \"edits\"}\`. Each file is processed independently in parallel — a hash mismatch on one file no longer blocks its siblings.
- Per-file results render as Markdown sections (\`## <path>\`) joined by \`---\` separators. Returns \`isError\` only when every file failed.
- Cap of 20 files per call, surfaced both in the JSON Schema (\`maxItems\`) and at runtime.

## Why the schema break is safe (per your #1 review item)

MCP serves \`tools/list\` dynamically on every session; clients always get the current \`inputSchema\`. The LLM sees the new shape on next connect and adapts — there is no cached-schema problem to migrate around. Calling this out explicitly so the safety argument doesn't have to be reconstructed during review. No version bump in this PR; happy to bump separately when you decide.

## Review fixes from #61 layered in

- **#4 (duplicate paths → data corruption)** — \`detect_duplicate_paths\` rejects the whole batch up front when two \`Ready\` tasks resolve to the same canonical path. Falls back to the literal path for files that don't exist yet, so the dedup key is still well-defined. The schema description and \`EDIT_MODE_EXTRA\` instructions both call out the unique-path requirement so the LLM groups edits correctly.
- **Medium (parse_file_edit short-circuit)** — \`parse_edit_entry\` returns the failing edit index in its error message, so the LLM can fix exactly the right entry instead of guessing which edit was malformed.
- **Medium (tests use tempfile::tempdir)** — new batch tests use \`tempfile::tempdir\` rather than fixed-name paths in \`std::env::temp_dir\`.

## Implementation shape

- \`src/edit.rs\` — \`pub apply_batch\` + \`pub FileEditTask::{Ready, ParseError}\`, with \`apply_one\` / \`render_applied\` keeping the parallel closure trivial. \`apply_edits\` and \`EditResult\` are now private — callers go through \`apply_batch\`.
- \`src/mcp.rs\` — \`tool_edit\` shrinks from ~85 lines of inline JSON parsing to: parse → cap-check → dedup-check → \`record_read\` → \`apply_batch\`. JSON parsing stays at the MCP wire boundary; parallel I/O and blast radius live next to \`apply_edits\`.

## Test plan

- [x] \`cargo build\` — clean
- [x] \`cargo test\` — 348 pass (4 new in \`edit::tests\` covering two-file success, partial failure, all-failed, and parse-error surfacing — all using \`tempfile::tempdir\`)
- [x] \`cargo clippy -- -D warnings\` — clean
- [x] \`cargo fmt --check\` — clean

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -60,7 +60,9 @@ Edit forms inside `edits`:
   Range:       {"start": "<line>:<hash>", "end": "<line>:<hash>", "content": "..."}
   Delete:      {"start": "<line>:<hash>", "content": ""}
 Per-file results: each file is processed independently. A hash mismatch on one file does NOT block the others.
+Response is `isError: false` whenever ≥1 file succeeded — always scan the per-file `## <path>` sections for failures rather than trusting the top-level status.
 Hash mismatch → file changed, re-read THAT file and retry it (other files in the batch already applied).
+A parse error on one edit invalidates ALL edits for that file (none applied); retry the whole file's edits after fixing the malformed entry.
 Each file path may appear at most once per call — group all edits for a file under its single entry.
 Large files: tilth_read shows outline — use section to get hashlined content.
 Pass diff: true to see a compact before/after diff per file.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -50,13 +50,19 @@ To find files, use tilth_files instead of Glob or Bash(find/ls).
 To check what changed, use tilth_diff instead of Bash(git diff/git log).
 DO NOT re-read files already shown in expanded search results.
 
-tilth_edit: Edit files using hash-anchored lines. Replaces the host Edit tool.
+tilth_edit: Batch edit one or more files using hash-anchored lines. Replaces the host Edit tool.
+ALWAYS group edits to multiple files into ONE tilth_edit call (max 20 files). Never call tilth_edit twice in a row.
 tilth_read → copy anchors (<line>:<hash>) → pass to tilth_edit.
-Single line: {"start": "<line>:<hash>", "content": "<new code>"}
-Range: {"start": "<line>:<hash>", "end": "<line>:<hash>", "content": "..."}
-Delete: {"start": "<line>:<hash>", "content": ""}
-Hash mismatch → file changed, re-read and retry.
+Shape: {"files": [{"path": "a.rs", "edits": [...]}, {"path": "b.rs", "edits": [...]}]}
+Single file: {"files": [{"path": "a.rs", "edits": [{"start": "<line>:<hash>", "content": "<new code>"}]}]}
+Edit forms inside `edits`:
+  Single line: {"start": "<line>:<hash>", "content": "<new code>"}
+  Range:       {"start": "<line>:<hash>", "end": "<line>:<hash>", "content": "..."}
+  Delete:      {"start": "<line>:<hash>", "content": ""}
+Per-file results: each file is processed independently. A hash mismatch on one file does NOT block the others.
+Hash mismatch → file changed, re-read THAT file and retry it (other files in the batch already applied).
+Each file path may appear at most once per call — group all edits for a file under its single entry.
 Large files: tilth_read shows outline — use section to get hashlined content.
-Pass diff: true to see a compact before/after diff in the response.
+Pass diff: true to see a compact before/after diff per file.
 After editing a function signature, tilth_edit shows callers that may need updating.
 DO NOT use the host Edit tool. Use tilth_edit for all edits.
--- a/src/edit.rs
+++ b/src/edit.rs
@@ -310,6 +310,46 @@ fn format_diffs(diffs: &[EditDiff]) -> String {
     out
 }
 
+/// Build a stable dedup key for a path. Canonicalise first (resolves symlinks
+/// and `.`/`..` when the file exists), fall back to `std::path::absolute`
+/// (resolves `.`/`..` and joins relative paths against cwd without touching
+/// the filesystem — catches not-yet-created aliases like `new.rs` vs
+/// `./new.rs`), then to the raw path. On macOS (commonly case-insensitive
+/// APFS) the key is ASCII-lowercased so `Foo.rs` and `FOO.RS` collide;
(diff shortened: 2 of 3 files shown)


jahala:
Reviewed by Opus 4.7:

The schema-break analysis is correct — verified `tools/list` is dispatch-time at `src/mcp.rs:360-367`, no client-side caching, so the new shape is picked up on next session. The parallel rayon apply is the right shape, and the `## <path>` Markdown sections + `---` separators is a clean rendering choice for the LLM to scan. One robustness item to address before merge, plus a few smaller things worth considering.

### Path-equivalence dedup for nonexistent files

```rust
// detect_duplicate_paths
let key = std::fs::canonicalize(path).unwrap_or_else(|_| path.clone());
```

For two batch entries with *exactly equal* path strings, this dedupes correctly via the raw-path fallback. The narrow case it misses: two entries with *different string spellings* of the same not-yet-existing file — e.g., `"new.rs"` and `"./new.rs"`, or on macOS `"new.rs"` and `"NEW.RS"` (case-insensitive APFS resolves to the same file). canonicalize fails for both (file doesn't exist), so each falls back to its own raw `PathBuf`, dedup misses them, and two rayon workers race on `fs::write` — last writer wins.

How likely is an LLM to do this? Low — they tend to pick one form and stick with it. But it's a silent data-loss class of bug if it does happen, so worth hardening.

Cheapest fix: lexically resolve the fallback before insertion.

```rust
let key = std::fs::canonicalize(path)
    .or_else(|_| std::path::absolute(path))
    .unwrap_or_else(|_| path.clone());
```

`std::path::absolute()` (stable in 1.79) joins relative paths against cwd and resolves `.`/`..` without touching the filesystem. A regression test with `"new.rs"` + `"./new.rs"` (neither exists) locks the contract.

### Smaller items (worth a follow-up, not blocking)

**Doc gap: single-edit failure aborts whole file.** `parse_file_edit` returns `ParseError` for the entire file on the first malformed `parse_edit_entry`. Reasonable conservatism, but the LLM will retry the whole file including the already-valid edits. A one-liner in `AGENTS.md` so the LLM knows what to expect on retry would help.

**Doc gap: partial success returns `isError: false`.** When 1 of 20 succeeds and 19 fail, the response is `Ok` because `any_success = true`. Defensible default — `isError: true` would risk the LLM retrying already-applied edits — but the schema description should call it out so the agent knows to scan per-file sections rather than trust the top-level success.

**Pre-existing carry-forward: `session.record_read` overcounts on partial failures.** The single-file path also called `record_read` before `apply_edits`, so a file that fails hash verification was still counted as "read." Multi-file scope makes it more visible. Not introduced by this PR, but a follow-up to gate `record_read` on `apply_one`'s success would tighten this.

**Test gaps:**
- Two tasks targeting the same nonexistent file under different spellings (proves the dedup fix above).
- One task with empty `edits: []` — should the `## <path>` section emit a non-empty body? Up to you, but worth pinning whichever choice you make.

### Nits

- `apply_one` takes `bloom: &Arc<BloomFilterCache>` and forwards to `render_applied`. `&Arc<T>` is double-indirection; either pass `Arc<T>` by value (workers clone) or `&BloomFilterCache`. Style only.
- The `fs::canonicalize(path).unwrap_or_else(|_| path.to_path_buf())` pattern appears in both `detect_duplicate_paths` and the per-task setup. A small `normalize_path_key` helper deduplicates.

Once the path-equivalence fix is in, this is good to merge. Want me to push it to your branch via `maintainerCanModify`, or would you rather keep it yourself?

paulnsorensen:
@jahala can you take a look at these when you get a chance? I have a lot more optimizations I'd like to contribute with the ultimate goal of batch writing/reading/searching to save on turns/tokens :)

jahala:
Hey @paulnsorensen  - thank you so much for your contributions! 🙏🏻 Really curious to see the next optimisations!

Merging now, and will retag and release later today (have to update a token) 

Here is what Opus 4.7 has to say

---

Round 2 looks great — the `normalize_path_key` extraction goes past what I'd asked for and the macOS handling is the right call. Specifically:

- The doc-comment on `normalize_path_key` names the trade-off explicitly ("false-positive collisions on case-sensitive APFS configs are preferred over false-negatives that race two writers against the same inode") — that's the load-bearing reasoning future readers will need.
- The three new tests cover the case I worried about (`new.rs` / `./new.rs` aliases on a nonexistent file), the don't-over-dedupe case (distinct nonexistent paths), and the macOS-specific case-insensitive collision. Pinned correctly.
- The helper is `pub(crate)` and lives in `src/edit.rs` next to the code that depends on the invariant — right module, right visibility.

Going to merge now. The smaller doc-gap items (single-edit abort, partial-success `isError`, `record_read` overcount) can land as follow-ups whenever you get to them — none are blocking.

On the broader direction: yes, batch writing/reading/searching is exactly the right next layer. The whole point of tilth is fewer round-trips per agent task; the benchmark data on v0.8.1 already shows roughly half-cost-per-correct vs baseline (haiku, ~74% volume-weighted on a 21-task slice), and batched MCP shapes are the most direct path to amplifying that further. Two things to keep in mind as you design more of them:

1. **Schema-break safety**: same argument as this PR's `files: [{path, edits}]` swap — fine because MCP's `tools/list` is dispatch-time, no client-side caching. Future batched shapes can keep adopting array-of-{...} freely.
2. **Hidden-shared-state hazards**: anywhere you fan out work in parallel, audit the path-key / cache-key invariant the way you did here. Most "batch X" PRs I've seen elsewhere ship a similar dedup race on the first try.

Bring on the next batches.

PR #81, opened 2026-05-09
refactor: adopt thiserror, fastbloom, and strsim

paulnsorensen:
## Summary

NIH audit findings — replace hand-rolled implementations with maintained crates.

- **`thiserror` v2** in `error.rs` — derive `Display`/`Error`, drops 25 lines of mechanical `write!` formatting.
- **`fastbloom` v0.17** in `index/bloom.rs` — drop hand-rolled `BloomFilter` (~105 LOC: `double_hash`, `combined_hash`, `hash_with_seed`, optimal-bit math). `BloomFilterCache` wraps `fastbloom::BloomFilter`, which is SIMD-optimized and the de facto Rust crate. The `extract_identifiers` byte state machine (the unique-to-tilth piece) stays.
- **`strsim` v0.11** in `diff/matching.rs` — replace `jaccard_similarity` (whitespace-set Jaccard) with `strsim::sorensen_dice` (character-bigram). Same `0.0..=1.0` range, same `0.8` fuzzy threshold.

Net: **+165 / -191** across 5 files (incl. Cargo.lock).

## Test plan

- [x] `cargo build` clean
- [x] `cargo clippy -- -D warnings` clean
- [x] `cargo test` — 248 passed (incl. all bloom + matching tests, including `fuzzy_match` and `below_fuzzy_threshold` which exercise the threshold)
- [x] `cargo fmt --check` clean

## Notes

- Drop `test_bloom_filter_sizing` — asserted private fields (`num_bits`, `num_hashes`) of the removed hand-rolled `BloomFilter` type. Equivalent guarantees are now part of fastbloom's own test suite.
- Sørensen-Dice on character bigrams is semantically distinct from token Jaccard, but the existing `fuzzy_match`/`below_fuzzy_threshold` tests pass at the same `0.8` threshold — the prefilter (size-ratio < 0.8 skips) keeps the candidate set small enough that bigram similarity behaves comparably for rename detection.

PR #87, opened 2026-05-09
refactor(bloom): adopt fastbloom for BloomFilter implementation

paulnsorensen:
## Summary

NIH audit finding — drop ~105 LOC of hand-rolled bloom-filter math in `src/index/bloom.rs` (`double_hash`, `combined_hash`, `hash_with_seed`, optimal-bit/hash sizing) in favor of [`fastbloom`](https://crates.io/crates/fastbloom), the SIMD-optimized de facto Rust bloom crate.

The `extract_identifiers` byte state machine — the unique-to-tilth piece — stays. `BloomFilterCache` now wraps `fastbloom::BloomFilter`, constructed via `BloomFilter::with_false_pos(0.01).expected_items(n)`.

Net: **+5 / -133** in `src/index/bloom.rs`; one new dependency (`fastbloom = "0.17"`).

This is one of three split PRs extracted from the bundled NIH audit refactor (see closed #81 / paulnsorensen/tilth#9). The other two — `thiserror` (#86) and `strsim` — are independent and land separately.

## Dependency health

`fastbloom` verified at time of writing (2026-05-08):

| Crate | Latest | Last release | Repo last push | Stars | Recent downloads |
|---|---|---|---|---|---|
| `fastbloom` | 0.17.0 | 2026-03-01 | 2026-03-01 | 348 | 5.0M |

Small project (348 stars), single maintainer, but very active: 0.15 → 0.16 → 0.17 across three weeks in Feb–Mar 2026. Only 4 open issues. The de facto fast Rust bloom crate; alternatives (`bloomfilter`, `growable-bloom-filter`) are unmaintained or niche. Non-archived, MIT/Apache-2.0.

## Test plan

- [x] `cargo build` clean
- [x] `cargo clippy -- -D warnings` clean
- [x] `cargo fmt --check` clean
- [x] `cargo test` — 342 passed (3 suites, all bloom tests including identifier extraction, FPR, and large-scale insertion)
- [x] Branch is off latest `origin/main` (post-v0.8.0)

## Notes

- Drop `test_bloom_filter_sizing` — it asserted private fields (`num_bits`, `num_hashes`) of the removed hand-rolled `BloomFilter` type. Equivalent guarantees are part of fastbloom's own test suite.
- `BloomFilter::new(expected_items, target_fpr)` (old, two-positional-arg constructor) becomes `BloomFilter::with_false_pos(0.01).expected_items(n)` (fastbloom builder pattern).

## Why split from #81

#81 bundled three independent NIH-audit changes. Reviewing them separately is easier — `fastbloom` is the largest and most controversial of the three because of single-maintainer dependency risk, so it deserves its own review thread.

The three split PRs touch disjoint files (`error.rs` / `index/bloom.rs` / `diff/matching.rs`) and disjoint `Cargo.toml` lines, so they merge in any order without conflict.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -235,6 +235,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "fastbloom"
+version = "0.17.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ef975e30683b2d965054bb0a836f8973857c4ebf6acf274fe46617cd285060d8"
+dependencies = [
+ "foldhash 0.2.0",
+ "libm",
+ "portable-atomic",
+ "rand",
+ "siphasher",
+]
+
+[[package]]
 name = "fastrand"
 version = "2.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -253,6 +266,24 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d9c4f5dac5e15c24eb999c26181a6ca40b39fe946cbe4c263c7209467bc83af2"
 
 [[package]]
+name = "foldhash"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "77ce24cb58228fbb8aa041425bb1050850ac19177686ea6e0f41a70416f56fdb"
+
+[[package]]
+name = "getrandom"
+version = "0.3.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "899def5c37c4fd7b2664648c28120ecec138e4d395b459e5ca34f9cce2dd77fd"
+dependencies = [
+ "cfg-if",
+ "libc",
+ "r-efi 5.3.0",
+ "wasip2",
+]
+
+[[package]]
 name = "getrandom"
 version = "0.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -260,7 +291,7 @@ checksum = "0de51e6874e94e7bf76d726fc5d13ba782deca734ff60d5bb2fb2607c7406555"
 dependencies = [
  "cfg-if",
  "libc",
- "r-efi",
+ "r-efi 6.0.0",
  "wasip2",
  "wasip3",
 ]
@@ -327,7 +358,7 @@ version = "0.15.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9229cfe53dfd69f0609a49f65461bd93001ea1ef889cd5529dd176593f5338a1"
 dependencies = [
- "foldhash",
+ "foldhash 0.1.5",
 ]
 
 [[package]]
@@ -410,6 +441,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b5b646652bf6661599e1da8901b3b9522896f01e736bad5f723fe7a3a27f899d"
 
 [[package]]
+name = "libm"
+version = "0.2.16"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b6d2cec3eae94f9f509c767b45932f1ada8350c4bdb85af2fcab4a3c14807981"
+
+[[package]]
 name = "linux-raw-sys"
 version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


jahala:
Reviewed by Opus 4.7:

Behavior parity confirmed — FPR target 0.01 preserved, builder-pattern equivalent to the old constructor, per-process random seed is fine for an in-memory cache keyed on path+mtime. The math is right and the swap is clean.

### Test count discrepancy — confirm before merge

Your PR description says "342 pass (down from 344)". Running `cargo test` on the checked-out branch shows **338**. Diff has two explicit deletions visible (`test_bloom_filter_sizing` and the constructor test), but the actual count is down 6, not 2.

Most likely explanation: the old hand-rolled `BloomFilter` struct had `#[cfg(test)] mod tests { ... }` content that disappeared along with the module, so the count delta includes those internal tests as well as the two visible top-level ones. That's correct behavior — those tests asserted on `num_bits` / `num_hashes` private fields that no longer exist on the opaque `fastbloom::BloomFilter`, and the equivalent guarantees are part of fastbloom's own test suite.

Could you confirm and update the PR description? A quick `git diff main -- src/index/bloom.rs | grep -c "^+.*#\[test\]"` vs the same for `^-` would settle it. Just want to make sure the count drop isn't masking a regression somewhere unrelated.

### Nit: lost context comment on filter sizing

`build_filter` now does:

```rust
let idents: Vec<&str> = extract_identifiers(content).collect();
let expected = idents.len().max(1);
let mut filter = BloomFilter::with_false_pos(0.01).expected_items(expected);
for ident in idents {
    filter.insert(ident);
}
```

`idents` contains many duplicates — same identifier (`fn`, `let`, `use`, `i`, `j`) appears repeatedly in any source file. So `expected_items(idents.len())` over-sizes the filter relative to actual unique-identifier load, which means achieved FPR is much better than 1% in practice. That's fine (the filter is just slightly bigger than necessary), but the deleted comment "Count identifiers first to size the filter appropriately" was load-bearing context for a future reader. Restore something like `// Sized for total token count, which over-allocates given duplicates — achieved FPR is < 0.01 in practice.`

Once the test count is verified and the comment is restored, this is good to merge.

jahala:
Reviewed by Opus 4.7:

Behavior parity confirmed — FPR target 0.01 preserved, builder-pattern equivalent to the old constructor, per-process random seed is fine for an in-memory cache keyed on path+mtime. The math is right and the swap is clean.

### Test count discrepancy — confirm before merge

Your PR description says "342 pass (down from 344)". Running `cargo test` on the checked-out branch shows **338**. Diff has two explicit deletions visible (`test_bloom_filter_sizing` and the constructor test), but the actual count is down 6, not 2.

Most likely explanation: the old hand-rolled `BloomFilter` struct had `#[cfg(test)] mod tests { ... }` content that disappeared along with the module, so the count delta includes those internal tests as well as the two visible top-level ones. That's correct behavior — those tests asserted on `num_bits` / `num_hashes` private fields that no longer exist on the opaque `fastbloom::BloomFilter`, and the equivalent guarantees are part of fastbloom's own test suite.

Could you confirm and update the PR description? A quick `git diff main -- src/index/bloom.rs | grep -c "^+.*#\[test\]"` vs the same for `^-` would settle it. Just want to make sure the count drop isn't masking a regression somewhere unrelated.

### Nit: lost context comment on filter sizing

`build_filter` now does:

```rust
let idents: Vec<&str> = extract_identifiers(content).collect();
let expected = idents.len().max(1);
let mut filter = BloomFilter::with_false_pos(0.01).expected_items(expected);
for ident in idents {
    filter.insert(ident);
}
```

`idents` contains many duplicates — same identifier (`fn`, `let`, `use`, `i`, `j`) appears repeatedly in any source file. So `expected_items(idents.len())` over-sizes the filter relative to actual unique-identifier load, which means achieved FPR is much better than 1% in practice. That's fine (the filter is just slightly bigger than necessary), but the deleted comment "Count identifiers first to size the filter appropriately" was load-bearing context for a future reader. Restore something like `// Sized for total token count, which over-allocates given duplicates — achieved FPR is < 0.01 in practice.`

Once the test count is verified and the comment is restored, this is good to merge.

paulnsorensen:
Verified before pushing — the 342 in the PR description is the total across all three suites; the `338` you ran into is the unit-tests-only line:

| Suite       | `main` | This PR | Δ |
|-------------|------:|--------:|--:|
| Lib unit    |   339 |     338 | −1 |
| Integration |     4 |       4 |  0 |
| Doctests    |     1 |       0 | −1 |
| **Total**   | **344** | **342** | **−2** |

So the −2 is exactly: −1 `test_bloom_filter_sizing` (asserted on the now-gone `num_bits` / `num_hashes` private fields) and −1 the `BloomFilter::new` doctest (removed with the constructor it documented). No regression hiding in the delta — PR description's "342 (down from 344)" stands.

Restored the filter-sizing comment in `build_filter`:

```rust
// Sized for total token count, not unique identifiers -- duplicates over-allocate
// the filter, so the achieved FPR is well below the 0.01 target in practice.
```

Also rebased onto latest `main` to clear the conflict with #86's `thiserror` addition. Single commit, force-pushed: `3ff87ca`.

PR #89, opened 2026-05-09
fix(search): kind="symbol" no longer leaks comment/string hits

paulnsorensen:
## The bug

Searching for a Python class via `tilth_search(query="Foo", kind="symbol")` returned the literal comment line `# class Foo` as a usage match, alongside the real `class Foo:` definition.

Reproducer (a snippet I hit in a real Python project, reduced):

```python
# class Foo
class Foo:
    pass

foo = Foo()
```

On `main`, run from the project root with the snippet above as `foo.py`:

```
$ tilth Foo
# Search: "Foo" — 3 matches (1 definitions, 2 usages)
### foo.py:2-3 [definition]
→ [2]   class Foo:
### foo.py:5 [usage]
→ [5]   foo = Foo()
### foo.py:1 [usage]   ← BUG: this is a comment, not a usage
→ [1]   # class Foo
```

`find_usages` was running a ripgrep word-boundary pass with no AST filter, so any comment or string-literal token whose text matched the query name leaked into `kind="symbol"` results.

## The fix

Make `kind="symbol"` strict by default — declarations only, no comment/string hits. The strict path skips the usage-merge step in `symbol::search` so comment/string scan results never reach the result list.

Add `kind="any"` for callers who want the prior mixed behavior (symbol-name matches including comments, strings, and usages).

```
kind: "symbol" (default, declarations only)
    | "any"     (symbol-name matches incl. comments/strings)
    | "content" (strings/comments)
    | "callers" (call sites)
```

`SERVER_INSTRUCTIONS` (`src/mcp.rs`) and `AGENTS.md` updated together so MCP hosts and the user-facing copy stay in sync.

## Test

`src/search/symbol.rs::strict_mode_drops_comment_and_usage_matches` — on the fixture above, `kind="symbol"` returns the single class definition; `kind="any"` returns all three matches.

## Benchmark spot-check

`kind="symbol"` is the hottest tilth path, so I ran the two hardest tasks per `CLAUDE.md` (`rg_search_dispatch`, `rg_trait_implementors`) with haiku, 2 reps, `--modes tilth`:

| Task | Branch (strict) | main | Δ |
|---|---|---|---|
| `rg_search_dispatch` rep0 | \$0.1333 / 78s ✓ | \$0.1435 / 62s ✓ | -7% cost |
| `rg_search_dispatch` rep1 | \$0.1067 / 55s ✓ | \$0.1639 / 83s ✓ | -35% cost |
| `rg_trait_implementors` rep0 | \$0.0212 / 13s ✓ | \$0.0188 / 10s ✓ | +13% cost |
| `rg_trait_implementors` rep1 | \$0.0150 / 12s ✓ | \$0.0173 / 9s ✗ | -13% cost, +1 correct |
| **Correctness** | **4/4** | **3/4** | +1 |
| **Avg cost / attempt** | **\$0.069** | **\$0.086** | -20% |

No latency regression. Slight directional improvement, plausibly because strict results stop feeding comment-hit noise into the model's reasoning loop. Sample is small — directional only.

## Behavior change note

This flips the default behavior of the most-used tool. Per project early-development stance, no deprecation shim is included; callers who want the prior behavior pass `kind="any"`.


paulnsorensen:
Wrong repo — moving to fork.

Issue #65, opened 2026-04-26
I love the premise of `tilth`, but it's not adding value as long as `--full` still heavily truncates results


giladbarnea:
Hi! I've been looking for a tool like `tilth` forever. Excited to try it out.

I get that being token-economic is a high-priority design choice, which is great. But I want to be able to get all results, without truncation. I thought `--full` did this — but there is no real difference between using it and not using it:

```
❯ npx tilth overlay --full | wc -l
      45
❯ npx tilth overlay | wc -l
      37
```

And:

```
❯ npx tilth overlay --full | tail
## client/src/lib/quakeConsole.js:116,117 [2 usages in createOverlay]
  [104-113]    fn toggle
             function toggle()
→ [115-126]    fn createOverlay
             function createOverlay()
  [128-135]    fn createToggleButton

... and 29 more matches. Narrow with scope.

(~350 tokens)%
```

Experience is — Well, **since I specified `--full`, please show me those 29 extra matches.**

("Overlay" is a widespread concept in this specific repo.)

Thanks for this whole thing!

jahala:
Hey @giladbarnea -- Thank you for trying tilth, and for reporting this! 

I'm not really a dev - so I'm using Opus to help out -- please see the review below. I'll keep tinkering this week! 

---

Thanks for the kind words and the clear repro — you're right, this is a real bug, not just a UX papercut.

## What's actually happening

`--full` is wired through to the file-read path (`tilth path/to/file.rs --full` returns the whole file instead of an outline), but for symbol/text/regex search queries it's effectively a no-op. The 8-line difference you see between `tilth overlay` and `tilth overlay --full` is incidental formatting; neither path expands more matches or raises the truncation cap.

For symbol searches there are actually two separate caps stacked:

1. **Walker cap** (`EARLY_QUIT_THRESHOLD_USAGES = 30`) — the search stops scanning files after collecting ~30 raw usage matches.
2. **Display cap** (`MAX_MATCHES = 10`) — the result is then truncated to 10 grouped matches before formatting, with the rest summarized as "and N more matches."

Neither is currently controlled by any user-facing flag. The "Narrow with scope" hint is the only escape valve, and it's not great when (as in your case) you legitimately want to see every use of a concept across the whole repo.

## Workarounds in 0.6.3

[...]

PR #112, opened 2026-05-16
Fold fork MCP/search/read updates

paulnsorensen:
## Summary

Manual patch-series PR for the fork fold-in. The head branch is the top of the stack; the individual slices are linked below so reviewers can inspect each delta without native Graphite/GitHub stack support.

Base commit: `e6b4514` (`origin/main` when the stack was created).

## Patch Series

1. `docs(mcp): externalize prompt instructions`
   - Adds external MCP prompt files, markdownlint config, prompt regen script, and markdown CI.
   - Slice diff: https://github.com/paulnsorensen/tilth/compare/e6b4514...96c4379

2. `feat(search): add strict symbol search mode`
   - Adds `SymbolMode` and CLI `--kind=symbol|any`; default symbol search returns declarations only.
   - Slice diff: https://github.com/paulnsorensen/tilth/compare/96c4379...0c32ccf
   - Refs #65: this keeps the fork's search-surface work linked, but does not close the remaining --full cap issue.

3. `feat(edit): report parse errors after writes`
   - Adds post-edit tree-sitter parse checking and tracks only committed files in batch edit outcomes.
   - Slice diff: https://github.com/paulnsorensen/tilth/compare/0c32ccf...da193fe

4. `feat(read): add budgeted sections and read modes`
   - Adds shared section budget behavior, truncation metadata, `signature` and `stripped` read modes, and read schema reconciliation.
   - Slice diff: https://github.com/paulnsorensen/tilth/compare/da193fe...9b1243d

5. `feat(mcp): expose v2 tool surface`
   - Splits `src/mcp.rs` into `src/mcp/`, advertises the v2 tool surface, and adds batched search/read/list/write behavior.
   - Slice diff: https://github.com/paulnsorensen/tilth/compare/9b1243d...4da37de

6. `test(mcp): pin v2 tool descriptions`
   - Adds `insta` snapshot coverage for the externalized v2 tool descriptions.
   - Slice diff: https://github.com/paulnsorensen/tilth/compare/4da37de...b274bb8

7. `feat(mcp): six-item friction sweep -- 5 fixed, 1 N/A`
   - Adds testing-session fixes for batched read not-found reporting, symbol-miss routing, read batching, and v2 tool-description guidance.
   - Slice diff: https://github.com/paulnsorensen/tilth/compare/b274bb8...085b5f6

## Maintainer Notes

This is intentionally opened as one draft PR because native stacked PR submit is not enabled for either `jahala/tilth` or `paulnsorensen/tilth`. The first six slice branches are pushed to `paulnsorensen/tilth` with stable names, and slice 7 is the top commit on `paulnsorensen/fork-mcp-polish-snapshots`.

```bash
git fetch https://github.com/paulnsorensen/tilth.git \
  paulnsorensen/fork-prompts \
  paulnsorensen/fork-symbol-kind \
  paulnsorensen/fork-edit-parse-check \
  paulnsorensen/fork-read-budget-modes \
  paulnsorensen/fork-mcp-v2-surface \
  paulnsorensen/fork-mcp-polish-snapshots
```

Maintainers should feel free to rebase or force-push the PR branch if that is easier than asking for follow-up rebases.

## Test Plan

- `npx --yes markdownlint-cli prompts/ AGENTS.md --config .markdownlint.json`
- `cargo fmt --check`
- `cargo clippy -- -D warnings`
- `cargo test`


paulnsorensen:
@jahala sorry for the giant diff, if you have graphite, I can submit a stack that way. Unfortunately it doesn't look like gh stack is available quite yet.

These changes were modeled after some usage with woz code as well as doing some session analytics on my end. I had claude go through usages of tilth (since I use it on all my projects and at work) and suggest what wasn't clear, which parameters didn't get enough use. You can look at my fork for more details, but the main bullet point was that the MCP needs instructions to remind the LLMs (instead of prompts) each time it's used, so I refactored to do that. Additionally, as you know, my goal was to get the LLM to batch as much as possible. 

Happy to connect to go over it in more detail if you'd like.

paulnsorensen:
Also, I just ran this with multi-edit benchmarks
[benchmark_20260516_165249_sonnet_tidy.md](https://github.com/user-attachments/files/27863718/benchmark_20260516_165249_sonnet_tidy.md)


[benchmark_20260516_165249_sonnet.md](https://github.com/user-attachments/files/27863721/benchmark_20260516_165249_sonnet.md)

jahala:
## Review summary

**Recommendation: HOLD.** This is a substantial PR (51 files, +6932/-1669) that lands genuinely useful infrastructure (externalized prompts, batched tools, hash-anchored writes with auto-fix, post-edit parse checking), and the architecture of the v2 surface is sound. But there are concrete blockers around silent breaking changes and a regression of recently-merged work that need addressing before merge. I'd love to take this in slices — see "Splitting suggestion" at the bottom.

I reviewed each of the six commits independently (via four rust-pro / code-reviewer subagents) plus did integration checks against current `main`. Findings are grouped by severity below, with the per-slice depth at the end.

---

## Blockers

### B1. NFS guard regression — silently reverts PR #110
- `src/search/mod.rs` on main has `.same_file_system(true)` in two places (PR #110, commit e4cb24e). The PR branch has zero occurrences. Verified:
  ```
  $ grep -c same_file_system main:src/search/mod.rs       → 2
  $ grep -c same_file_system pr-112:src/search/mod.rs     → 0
  ```
- Because the PR base is `e6b4514` (pre-#110), the rebase will need to consciously re-introduce the guard. GitHub showing "MERGEABLE" is git-mergeable, not semantically-mergeable.

### B2. `kind="symbol"` is a silent breaking change for MCP callers
- Current main MCP schema (`src/mcp.rs:924-928`):
  ```
  "kind": { "default": "symbol", "description": "symbol: structural definitions + usages." }
  ```
- After this PR, `kind="symbol"` maps to the new `SymbolMode::Strict` = **definitions only**, no usages.
- Every existing agent doing `tilth_search(query="handleRequest")` with no explicit `kind` was getting defs + usages + identifier-shaped callers; it now gets just the definition. The PR description does not flag this.
- Either keep `"symbol"` meaning what it currently means and add a new value (`"definitions"` / `"strict"`) for the new behavior, or version-bump and document the break.

### B3. CLI `--kind=any` is not in the MCP enum
- The CLI accepts `--kind=any` (`CliKind::Any`), but the MCP `kind` enum is still `["symbol", "content", "regex", "callers"]`. An MCP caller passing `kind="any"` hits the `_ => Err(...)` arm and gets an unknown-kind error. CLI and MCP surfaces must agree.

### B4. Built-in tool removals with no deprecation path
- `tilth_files` and `tilth_edit` are removed from `tool_definitions` (snapshot test `tool_definitions_expose_v2_surface_only` pins this). No runtime alias.
- For users running `tilth install <host>` and then upgrading, mid-conversation Claude Code sessions that already have `tilth_files`/`tilth_edit` in their tool-list cache will fire unknown-tool errors until they reconnect. Worth at minimum a major version bump + clear release note; ideally a one-release alias that emits a deprecation notice in the response.

---

## Concerns

### C1. Prompt content was rewritten, not just externalized (slice 1)
- `prompts/mcp-base.md` replaces ~12 paragraphs of per-tool documentation in `SERVER_INSTRUCTIONS` with a four-bullet summary. `prompts/mcp-edit.md` drops the entire `tilth_edit` parameter block (request shape, per-file failure semantics, hash-mismatch retry, `diff: true`).
- Haiku is sensitive to instruction positioning + framing (CLAUDE.md explicitly notes this). The hard tasks (`rg_search_dispatch`, `rg_trait_implementors`, `gin_servehttp_flow`) need a haiku benchmark before merging this content change.
- The infrastructure (regen-agents-md.sh, markdownlint config, CI markdown job) is fine — it's the content swap that's risky.

### C2. CI has no AGENTS.md drift check
- `.github/workflows/ci.yml` lints `prompts/` and `AGENTS.md` separately, but does not run `scripts/regen-agents-md.sh && git diff --exit-code AGENTS.md`. Drift between prompts and AGENTS.md will accumulate silently. (And `AGENTS.md` is already unchanged in this PR even though prompts have changed — it's already diverged at PR-time.)

### C3. Go doc-comment stripping in `mode=stripped`
- `src/search/strip.rs:is_strippable_comment` for `StripLang::Go` is `trimmed.starts_with("//")` with no doc-comment exception. In Go, package-level godoc IS `// text` immediately above a declaration — there's no syntactic difference. So `mode=stripped` will silently destroy godoc on every Go file. Every other language has a `///` / `/**` / `"""` escape hatch.

### C4. Python `print()` stripped as "debug log"
- `mode=stripped` applies strip_noise to the whole file (not just expanded definitions, like the current search-expand path). Stripping `print()` in CLI scripts, test helpers, REPL output, etc. silently breaks comprehension. The stripped output is also labeled with `//` regardless of file language — Python files get `// stripped N of M lines`.

### C5. 3-hex hash = 12 bits → high collision risk
- `src/format.rs:line_hash` is FNV-1a truncated to 12 bits (4096 buckets). For a 5000-line file, pairwise birthday collision is ~83%. Used as a sanity check only (write path verifies hash at the claimed line number, not via scan), and auto-fix catches the stale case, so collisions don't cause wrong-line edits. But a 1-in-4096 false-positive means an incorrect edit can land silently in the case where: agent reads line N at hash H, user concurrently changes line N to different content that hashes to H, agent writes. Worth widening to 6 hex (24 bits) or documenting the bound.

### C6. `truncated_at_line` gaps in multi-section reads
- `read_ranges_with_budget` returns a raw string for multi-section omission-marker paths without going through `finalize_response`, so `truncated: true` / `truncated_at_line` never get set when sections are dropped. The doc contract in `prompts/tools/read.md:17` says those fields "appear when a `budget` arg actually clipped the body" — broken.
- For `mode=signature` / `mode=stripped`, the body's line numbering diverges from disk line numbering, but `truncated_at_line` is reported in body-line units — meaningless to the agent. No test covers this divergence.

### C7. Search batch fails-fast on first error
- `src/mcp/tools/search.rs:38-70` runs queries sequentially with `?` propagation. A bad `glob` on query 2 of 5 aborts queries 3-5. Per-query error isolation (push an error section, continue) would be more resilient. Read batch handles this correctly already — search should match.

### C8. `kind=symbol` Concept path silently uses `Any`
- `src/lib.rs:281` hard-codes `SymbolMode::Any` for `QueryType::Concept` even when the caller passed `--kind=symbol`. The user passes "strict", gets relaxed. Either honor the flag or document the override.

### C9. Parse-check Parser allocation per call
- `edit_parse_check::parse_errors` constructs a new `tree_sitter::Parser` + `set_language` on every call. `check` calls it twice per edit. On a 100-file `par_iter` batch: 200 parser allocs. Not architecturally consistent with `OutlineCache`. Probably <1ms in practice, but worth a thread-local parser cache.

### C10. Stripped-mode "non-editable" not enforced
- `prompts/tools/read.md` says stripped output "cannot round-trip through tilth_write." That's enforced only by documentation — nothing in `tool_write` rejects line numbers from a stripped read. Either guard at the tool boundary (mark the response as non-editable in metadata, refuse writes against stripped reads) or accept the doc-only guarantee explicitly.

---

## Nits

- `if_modified_since`: condition is `mtime > since`; within the same second a file written after a read returns stub on the next call. `mtime >= since` would be safer, at the cost of re-delivering when wall-clock is coarse.
- `if_modified_since` uses mtime, not content hash — touch bumps mtime and triggers re-delivery of identical content. Correct but mildly wasteful.
- Snapshot tests pin tool descriptions but not JSON-RPC wire path. One end-to-end test parsing a raw newline-delimited response would close the gap.
- `CliKind::Symbol → SymbolMode::Strict` naming mismatch is confusing in error messages.
- `include_str!` on an empty `prompts/mcp-base.md` compiles silently with an empty string. Compile-time `assert!(!STR.is_empty())` is a trivial guard.
- `src/mcp/mod.rs` is 2638 lines (mostly tests). If it keeps growing, split tests into `tests/mcp_*.rs`. Not blocking.

---

## Rebase complexity

**HIGH on the search subsystem.** The just-merged #65 cap fix (56c3351) added a `full: bool` parameter to `symbol::search`, `content::search`, `search_callers_expanded`, `find_callers_batch`, and `search_multi_symbol_expanded`. This PR independently adds a `mode: SymbolMode` parameter to the same `symbol::search` and its public wrappers, plus restructures `src/mcp.rs` into `src/mcp/`. Reconciliation needs:

1. `symbol::search(query, scope, context, glob, mode: SymbolMode, full: bool)` — both params present
2. The `Strict` arm in `search()` needs the `full`-derived `def_threshold` plumbed into `find_definitions` so `--full --kind=symbol` raises the definition cap
3. `find_callers_batch` and `search_callers_expanded` are now in `src/mcp/tools/...` — the #65 cap-fix wiring needs to re-land there
4. `src/search/mod.rs`: preserve `.same_file_system(true)` (PR #110) at the `WalkBuilder` construction inside whatever new search dispatch lands

GitHub "MERGEABLE" is misleading — these are semantic conflicts at the same function bodies, not just textual.

---

## Splitting suggestion

The PR is hard to land as one merge precisely because the six slices have very different risk profiles. If feasible, I'd take them as separate PRs:

1. **`docs(mcp): externalize prompt instructions`** — easy yes if the content swap is reverted (or restored to parity). Land the regen script + markdownlint + CI infrastructure first.
2. **`feat(edit): report parse errors after writes`** — close to ready. One doc sentence in `prompts/tools/write.md` clarifying that `── parse ──` blocks are warn-only, and a parser cache is a nice-to-have not a blocker. Land this independently.
3. **`feat(search): add strict symbol search mode`** — needs the `kind="symbol"` semantic decision before merge (B2). Once decided, this is small.
4. **`feat(read): add budgeted sections and read modes`** — needs Go godoc fix (C3), Python print() decision (C4), and `truncated_at_line` fixes for multi-section + signature/stripped (C6).
5. **`feat(mcp): expose v2 tool surface`** — biggest, needs the `tilth_files`/`tilth_edit` deprecation story (B4), version bump, and the prompt-content question resolved. Probably worth its own focused review pass.
6. **`test(mcp): pin v2 tool descriptions`** — trails slice 5.

Happy to do a second pass on any individual slice if you want. Thanks for the structured patch-series description — it made this much easier to review.

jahala:
## Follow-up: race conditions, TOCTOU, state, edge cases

Did a focused second pass on concurrency / state / TOCTOU / edge-case correctness, since the first review was mostly structural. Every claim below is verified by direct file reading (not just subagent assertion).

### HIGH — Path-traversal escape in `mode=hash`

`tool_write` enforces scope containment only for `overwrite` / `append` modes (`src/mcp/tools/write.rs:64-84`). The `hash` branch at line 86 pushes the task straight into `crate::edit::apply_batch` with no scope check. The inline comment at write.rs:60 claims hash mode is safe because "`edit::apply_batch` which canonicalizes + roots to `package_root`" — but tracing the call chain:

- `apply_batch` → `apply_one` → `render_applied` → `apply_edits` (`src/edit.rs:67`)
- `apply_edits` does `fs::read_to_string(path)` → verify hashes → `fs::write(path, &output)` (line 219). No path containment anywhere.
- `normalize_path_key` (line 331) does canonicalize, but only as a dedup key for `detect_duplicate_paths` — not for containment.
- The `canonicalize` at `edit.rs:464` runs *after* the write succeeded, for blast-radius scope computation.

Test coverage gap: `tool_write_overwrite_outside_scope_refused` (`src/mcp/mod.rs:2157`) tests only `mode=overwrite`. There is no equivalent test for hash mode escape.

Attack path: agent reads `/etc/hosts` via `tilth_read` (no scope check there either — pre-existing, but exploitable here), gets `<line>:<3-hex-hash>` anchors, then `tilth_write({path: "/etc/hosts", mode: "hash", edits: [...]})` writes outside the project root. The current `overwrite` test even uses `outside_dir.path()`, so the user clearly considered this attack surface — but missed the hash branch.

**Fix**: move the `path_within_scope` check above the `match mode` block at write.rs:53, or repeat it in the hash arm.

### HIGH — TOCTOU window in hash-anchored auto-fix

The auto-fix retry chain reads the file three (potentially four) times with no advisory locking:

1. `capture_hash_original` (`src/mcp/tools/write.rs:245`) → `fs::read_to_string` for the pre-image
2. `apply_edits` (`src/edit.rs:77`) → `fs::read_to_string` for the hash verify + write
3. `auto_fix_locate` (`src/mcp/write.rs`) → `fs::read_to_string` for the body-relocation scan
4. `reapply_at_relocation` → another `fs::read_to_string` to compute fresh hashes

Between (3) and (4), an external writer can mutate the file. The realistic damage modes:

- **Wrong-line write**: a concurrent partial write produces a file whose `\n`-split lines momentarily contain the needle body bytes at exactly one position by coincidence. `auto_fix_locate` returns `Relocated{new_line}`, `reapply_at_relocation` recomputes hashes against that partial-flush snapshot, the hashes happen to match. Edit lands at the wrong line. Probability is small but non-zero for short edit bodies.
- **Silent failed retry**: if the file shifts again between (3) and (4), the recomputed hashes mismatch, `reapply_at_relocation` returns `None`, and the agent gets a "re-apply failed" message. No data loss, but a confusing UX.

`tool_write_auto_fix_shift_returns_fresh_region_not_relocation` (mod.rs:2129) tests the *intended* shifted-body path with a synthetic file, but doesn't exercise the partial-write race. Mitigation requires either: (a) `flock(2)` advisory locks, (b) read-then-hold the original bytes through every phase rather than re-reading, or (c) accept the race and document it.

### MEDIUM — `if_modified_since` same-second race

`file_changed_since` (`src/mcp/iso.rs:117-119`):
```rust
match metadata.modified() { Ok(mtime) => mtime > since, ... }
```

Scenario: server returns response at T (`if_modified_since: ISO(T)`). External editor writes the file later in the same second — mtime = `T.subsec_nanos = N`. Agent passes the token back, server reads `since = ISO(T) = T.subsec_nanos = 0` (since `iso_ts` truncates to second resolution), checks `mtime(T.N) > since(T.0)` → true, returns fresh content. **OK on nanosecond filesystems.**

But on HFS+ (1-second mtime resolution) or FAT/SMB (2-second): mtime = exactly T, since = exactly T → `mtime > since` → false → `(unchanged @ T)` stub returned for content that did change. Agent edits stale content. Hash-anchored writes will catch this on the next round-trip via hash mismatch, but overwrite/append will silently land on stale content.

Fix is one of: `mtime >= since` (over-deliver, never miss); compare at full SystemTime precision (don't round-trip through ISO seconds); or document the coarse-fs limitation in `prompts/tools/read.md`.

### MEDIUM — `OutlineCache` stale-serve under same-second external edit

`OutlineCache` key is `(PathBuf, SystemTime)`. If tilth populates the cache for file A at mtime=T, then an external tool overwrites A within the same second on HFS+/FAT (mtime still = T), the next read hits the same cache key and serves the stale outline. tilth's own writes bump mtime forward, so this is specifically an external-writer race on coarse-mtime filesystems. Same mitigation family as the `if_modified_since` issue.

### LOW — Search batch fails-fast (verified)

`src/mcp/tools/search.rs:38-70` runs the query batch sequentially with `?` propagation. One bad `glob` in query 2 of 5 aborts queries 3-5. The read batch handles this correctly (per-file inline errors); search should match. Confirmed by reading both paths.

### LOW — Path-suffix parser edge cases

`src/mcp/path_suffix.rs` correctly handles `path#0`, `path#-5`, `path#1-0`, empty `path#` — all fall through to safe behavior (no panic, predictable result). The notable gap is `path` containing `#` (filenames like `src/foo#bar.rs`): the parser uses `spec.find('#')` (first occurrence), which truncates the path. No escape mechanism. Unix-rare but possible. Heading-vs-symbol disambiguation by `contains(' ')` mis-classifies the rare case of identifiers with spaces. Both are low-risk.

### LOW — Parser allocation in parse-check

`edit_parse_check::parse_errors` allocates a new `tree_sitter::Parser` per call (twice per `check` — once for pre, once for post). On a 100-file `par_iter` batch that's 200 parser allocations. Not architecturally consistent with `OutlineCache` thread-local reuse elsewhere. Probably <1ms per call in practice, but worth a thread-local cache if batches grow.

### SAFE — parallel walker atomics + duplicate-path dedup

`detect_duplicate_paths` (edit.rs:347) canonicalizes via `normalize_path_key` (case-folds on macOS) and rejects the whole batch before any worker starts. Two rayon writers cannot race the same inode. The PR introduces no new atomic counters; existing `Ordering::Relaxed` patterns in the walker are correct because `walker.run()` joins before reads. No new concurrency primitives to audit.

### SAFE — post-write parse check non-atomicity (already understood)

`apply_edits` writes the file, then parse-checks the in-memory `output` string (never re-reads from disk). The on-disk content and the parsed content are byte-identical except for the OS write buffer, which is a kernel-level invariant. If parse-check reports errors, the file stays edited — documented warn-only behavior. No correctness bug, but as flagged in the first review: `prompts/tools/write.md` does not document that `── parse ──` blocks mean "written but broken."

---

**Highest-severity issue: the hash-mode path traversal (HIGH).** It's a documented assumption that turns out to be wrong, and it's directly exploitable. The auto-fix TOCTOU is a real but narrower race. The two MEDIUM mtime issues are coarse-filesystem-specific and have correct fallback paths via hash mismatch — annoying, not dangerous.

jahala:
Closing the fold — with the observation that time largely proved its direction right. Of its major slices, main independently landed prompt externalization, the mcp module split, signature/stripped read modes, and batched multi-section reads; the ideas won even where this specific diff couldn't merge. What kept the PR itself unmergeable was the 7-commit fold shape (the HOLD's split proposal still stands as the lesson) plus the two behavior changes that needed evidence, and an 11-week gap main outran.

Two concrete follow-ups are planned rather than lost: (1) the post-write parse-error reporting slice — our failure analysis now shows ~12% of benchmark failures are edit-shaped, which makes that slice evidence-backed; it will come back as a small focused PR with Co-authored-by credit to you. (2) A genuine ask: this PR mentioned being informed by your cross-project session analytics. We just ran our own real-session study and it overturned one of our benchmark-based conclusions — if you're willing to share aggregate numbers from yours, they'd directly inform two open design decisions. That data is arguably the most valuable thing in this PR.

PR #85, opened 2026-05-09
Extract MCP prompts to markdown, improve clarity, untrack prek.toml

paulnsorensen:
## Summary

Three commits extracting and improving MCP prompt management:

1. **Extract MCP prompts to markdown** — Move `SERVER_INSTRUCTIONS` and `EDIT_MODE_EXTRA` from inline strings in `src/mcp.rs` to dedicated markdown files (`prompts/mcp-base.md`, `prompts/mcp-edit.md`). Wire them back with `include_str!()` at compile time. Benefits: clearer separation of concerns, easier editing/versioning, markdown tooling support.

2. **Improve prompt clarity** — Polish MCP instructions: better grammar, clearer examples, refined guidance for Haiku. Add markdown linter config to catch formatting issues early.

3. **Untrack prek.toml** — Move prek.toml to `.git/info/exclude` (local-only, not checked in). Reduces noise in `.gitignore`, respects local-only tool config.

## Changes

- `prompts/mcp-base.md` — new base MCP instructions (extracted from `SERVER_INSTRUCTIONS`)
- `prompts/mcp-edit.md` — new edit-mode instructions (extracted from `EDIT_MODE_EXTRA`)
- `scripts/regen-agents-md.sh` — new script to regenerate `AGENTS.md` from markdown source
- `.markdownlint.json` — new markdown linter config (enforces clarity + consistency)
- `.github/workflows/ci.yml` — added markdown lint check
- `src/mcp.rs` — refactored to wire prompts via `include_str!()`, reduced from 162 lines to 55 lines
- `AGENTS.md` — regenerated from markdown source (clearer examples, improved phrasing)
- `CLAUDE.md` — updated version reference

PR #122, opened 2026-05-19
refactor(mcp): extract server instructions to prompts/*.md

paulnsorensen:
## Summary

Extract the inline `SERVER_INSTRUCTIONS` and `EDIT_MODE_EXTRA` string constants in `src/mcp/mod.rs` into versionable markdown sources at `prompts/mcp-base.md` and `prompts/mcp-edit.md`, wired back at compile time via `include_str!`. `AGENTS.md` becomes a generated artifact regenerated by `scripts/regen-agents-md.sh`, with CI failing if anyone edits `prompts/` without re-running the script.

Runtime MCP `instructions` output is byte-identical to pre-refactor — locked by three new unit tests.

## Relationship to prior PRs

- **Supersedes #85** (closed: opened against the wrong fork target). Same intent, clean against this fork's `main`.
- **Re-shapes slice 1 of #112**, but using the current v1 tool surface so it stands alone on `origin/main`. The slice 1 commit in that stack (`96c4379`) added `prompts/*.md` content that described the v2 tool surface (`tilth_list`, `tilth_write`) which doesn't exist yet on `main` — lifting it standalone would have produced orphan docs and a regen script that actively corrupted `AGENTS.md`. This PR uses the v1 text verbatim so the extraction is a true no-op refactor and is decoupled from the rest of the #112 stack.

## What changes

- `+ prompts/mcp-base.md` (2952 B) — verbatim text of current `SERVER_INSTRUCTIONS`, no trailing newline
- `+ prompts/mcp-edit.md` (1724 B) — verbatim text of current `EDIT_MODE_EXTRA` (keeps the leading `\n\n` that `format!("{S}{E}")` relies on)
- `+ scripts/regen-agents-md.sh` — idempotent generator: header comment + `cat prompts/mcp-base.md` + `cat prompts/mcp-edit.md` + final newline
- `+ .markdownlint.json` — disables MD012/13/22/33/40/41/47 (rules that conflict with the verbatim-from-source markdown shape)
- `~ src/mcp/mod.rs` — const bodies replaced with `include_str!`, doc comment updated, three new `#[cfg(test)]` tests:
  - `server_instructions_byte_lock` — locks byte count (2952), prefix, suffix, no triple newlines, two interior content markers
  - `edit_mode_extra_byte_lock` — same shape (1724 bytes); also asserts the leading `\n\n` is preserved
  - `instructions_compose_with_single_blank_line_between_sections` — pins the exact `"results.\n\ntilth_edit:"` boundary so the runtime composition shape is enforced
- `~ AGENTS.md` — regenerated from the prompts/ sources; absorbs the pre-existing `SERVER_INSTRUCTIONS ↔ AGENTS.md` drift by canonicalising on `SERVER_INSTRUCTIONS` (the runtime-visible copy wins)
- `~ .github/workflows/ci.yml` — new `markdown` job runs SHA-pinned `markdownlint-cli` over `prompts/` and `AGENTS.md`, plus a freshness step that fails if `bash ./scripts/regen-agents-md.sh && git diff --exit-code AGENTS.md` is non-empty
- `~ CLAUDE.md`, `~ ARCHITECTURE.md` — project-structure entries and MCP-server descriptions reflect the new layout

## Pre-existing drift, reconciled

`SERVER_INSTRUCTIONS` and `AGENTS.md` had already diverged on at least five lines (multi-symbol-lookup phrasing, `glob` examples, session-dedup line, `tilth_files` output, edit anchor language). `CLAUDE.md` declared they "should stay in sync" but enforcement was manual. This PR canonicalises on `SERVER_INSTRUCTIONS` so the runtime `instructions` string MCP hosts receive is byte-identical to pre-refactor; the human-readable `AGENTS.md` gets the runtime canon. From here, the regen script + CI freshness gate makes future drift impossible to land.

If you'd prefer to keep `AGENTS.md` hand-edited as the source of truth, the fallback is small: flip the regen script into a `--check` mode and require updates to `AGENTS.md` to be reflected in `SERVER_INSTRUCTIONS`. Happy to make that adjustment if that's the preferred direction.

## Out of scope (explicit)

- v2 tool surface (`tilth_list`, `tilth_write`) — slice 5 of #112, separate PR
- Prompt rewording / Haiku tuning — needs benchmark validation, separate PR
- Slices 2–7 of #112 — independent extractions on follow-up branches
- `insta` snapshot migration — slice 6 of #112; this PR uses literal `assert!` / `assert_eq!` to avoid adding a new dependency

## Test plan

- `cargo test --release` — **369 passed** (3 new byte-lock tests)
- `cargo fmt --check` — clean
- `cargo clippy -- -D warnings` (CI form) — clean
- `npx markdownlint-cli prompts/ AGENTS.md --config .markdownlint.json` — ok
- `bash ./scripts/regen-agents-md.sh && diff` against pre-regen snapshot — verified no-op (the CI freshness gate will pass on a clean checkout)
- Manual: `cargo build --release` produces a binary whose MCP `initialize` `instructions` field is byte-identical to pre-refactor (enforced transitively by the const byte locks + unchanged `format!()` call sites at `src/mcp/mod.rs:306-313`)

🧀 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,3 +23,17 @@ jobs:
       - run: cargo fmt --check
       - run: cargo clippy -- -D warnings
       - run: cargo test
+
+  markdown:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # ratchet:actions/checkout@v6
+      - uses: nosborn/github-action-markdown-cli@9b5e871c11cc0649c5ac2526af22e23525fa344d # ratchet:nosborn/github-action-markdown-cli@v3.3.0
+        with:
+          files: prompts/ AGENTS.md
+          config_file: .markdownlint.json
+      - name: AGENTS.md is regenerated from prompts/
+        # Re-running the regen script must be a no-op. If AGENTS.md drifts from
+        # prompts/*.md (someone edited the markdown but forgot to re-run the
+        # script), this step fails and surfaces the missing regen in the PR.
+        run: bash ./scripts/regen-agents-md.sh && git diff --exit-code AGENTS.md
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,0 +1,10 @@
+{
+  "default": true,
+  "MD013": false,
+  "MD041": false,
+  "MD012": false,
+  "MD047": false,
+  "MD033": false,
+  "MD022": false,
+  "MD040": false
+}
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -160,9 +160,10 @@ methods worth describing in detail:
 
 - `initialize` — emits `protocolVersion`, capabilities, `serverInfo`,
   and an `instructions` string. The instructions are the
-  `SERVER_INSTRUCTIONS` constant (recently rewritten on `fd3de77` as a
-  pre-flight gate naming the exact Bash commands and host tools to
-  avoid, with concrete `<bad>→<good>` rewrites). When
+  `SERVER_INSTRUCTIONS` constant, loaded at compile time via
+  `include_str!("../../prompts/mcp-base.md")` and rewritten on `fd3de77`
+  as a pre-flight gate naming the exact Bash commands and host tools to
+  avoid, with concrete `<bad>→<good>` rewrites. When
   `TILTH_NO_OVERVIEW` is unset, `overview::fingerprint(cwd)` is
   prepended — a project summary built in <250ms (a stderr warning fires
   if it overruns).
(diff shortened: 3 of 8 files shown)


jahala (approved):
## Verdict
**APPROVE** (one optional nit).

Did a deep independent review: verified the byte-identical claim, ran build/test/clippy/fmt locally, inspected the embed mechanism, and confirmed the CI freshness gate works as advertised. The PR is correctly scoped and the infrastructure is solid.

## Verified byte-identity
Compiled the pre-refactor `SERVER_INSTRUCTIONS` (from `main:src/mcp.rs:58`) and `EDIT_MODE_EXTRA` (`main:src/mcp.rs:112`) as standalone strings and measured `.len()`:

| Constant | Pre-refactor bytes | `prompts/*.md` size | Match |
| --- | --- | --- | --- |
| `SERVER_INSTRUCTIONS` | 2952 | 2952 | ✅ |
| `EDIT_MODE_EXTRA` | 1724 | 1724 | ✅ |

Worth flagging for future reviewers (and future me): the visual indentation in the original source string literals (`  For multi-symbol lookup, ...`) was being **eaten by Rust's `\<newline>` line-continuation** before reaching the runtime — that's `unicode-string-continue` semantics, which strips the `\n`, any subsequent whitespace, and the `\` itself. So the runtime string never had the indentation the source code visually implies. The `prompts/*.md` files match what was actually being sent over MCP. The three byte-lock tests (`server_instructions_byte_lock`, `edit_mode_extra_byte_lock`, `instructions_compose_with_single_blank_line_between_sections`) pin this correctly.

## Verified locally
- `cargo build --release` — clean (23.10s)
- `cargo test --release` — **369 passed, 0 failed** (incl. the 3 new byte-lock tests)
- `cargo clippy --release -- -D warnings` — zero warnings
- `cargo fmt --check` — clean
- `include_str!` mechanism confirms binary self-contained — no runtime file reads
- Freshness gate (`bash ./scripts/regen-agents-md.sh && git diff --exit-code AGENTS.md`) verified present in `.github/workflows/ci.yml`

## What I like
- The "canonicalize on `SERVER_INSTRUCTIONS` because it's the runtime-visible copy" call is correct — that's the string benchmarks have been tuned against. AGENTS.md absorbing the pre-existing drift is the right direction.
- Three orthogonal byte-lock tests (length, prefix/suffix, composition boundary) form a reasonable safety net against accidental edits to `prompts/*.md`.
- Freshness gate + markdownlint job give us automated enforcement of what `CLAUDE.md` previously stated as a manual convention.
- Out-of-scope section is honest: v2 surface, prompt rewording, slices 2–7 of #112 explicitly deferred. No silent piggybacking.

## LOW / nit (optional)
- `ARCHITECTURE.md` references the bare commit hash `fd3de77` as context for where the constants used to live. That'll go stale; consider replacing with "pre-refactor `src/mcp.rs`" or "before #122" so future readers don't need to git-archaeology.

## Recommendation
Merge. No reservations on the substance.

PR #124, opened 2026-05-20
feat(edit): allow tilth_edit to create new files (closes #123)

jahala:
Closes #123.

## Summary

`tilth_edit` can now create new files via a `create: true` flag + `content` at the file level (mutually exclusive with `edits`):

\`\`\`json
{
  "files": [
    {"path": "existing.rs", "edits": [{"start": "42:a3f", "content": "..."}]},
    {"path": "new.rs", "create": true, "content": "fn main() {}\n"}
  ]
}
\`\`\`

## Design

- **File-level discriminator**, not edit-level — a create is a file-level operation, no edit-array semantics
- **Atomic** via `OpenOptions::create_new(true)` — O_CREAT|O_EXCL on Linux/macOS, CREATE_NEW on Windows. No TOCTOU window, no silent overwrite
- **Parent dirs auto-created** via `fs::create_dir_all` (idempotent under concurrent creates)
- **Symlinks not followed** — dangling symlink at target → `AlreadyExists` (verified by unix-gated test)
- **Hashlined output** of the new file returned, so follow-up edits work without re-reading

## Internal API

- New `FileEditTask::Create { path, content }` variant — `Edit` struct unchanged
- New `TilthError::AlreadyExists` variant — typed error class for clearer agent UX
- `detect_duplicate_paths` extended to cover Create
- `Session::record_read` tracks Create paths (creating a file is higher-commitment than reading one)

## Test coverage

16 new tests across three layers:
- **create_file unit (7)**: happy path, exists-already, parent-dir auto-create, empty content (touch), hashlined output for follow-up, permission denied (cfg(unix)), dangling symlink (cfg(unix))
- **apply_batch integration (3)**: mixed create+edit batch, one-fails-others-succeed, duplicate Create paths rejected up front
- **parser (6)**: create+content → Create, create+edits → ParseError, missing content → ParseError, stray content → ParseError, create:false is edit-shape, string \"true\" rejected (strict bool)

## Audit

Second-pass audit by code-reviewer subagent found 2 HIGH (empty-create returned header-only response; Create paths skipped in session), 4 MEDIUM (HashMismatch arm should be unreachable; doc comment drift; function-body imports; missing symlink test), and 3 LOW (typed AlreadyExists variant; cfg attribute on platform-skip test; render_created blast-radius comment). All fixed before commit — full audit log in commit message.

## Test plan

- [x] \`cargo build --release\` clean
- [x] \`cargo test --release\` — 412 passed, 0 failed (+ 16 new)
[...]

PR #126, opened 2026-05-23
feat(edit): post-write parse check for tilth_edit

paulnsorensen:
## What

Adds a **post-write parse check** to `tilth_edit`. After an edit writes a file, the pre- and post-edit content are parsed with the language's tree-sitter grammar, and a `── parse ──` block is appended to the tool output listing only the **new** `ERROR` / `MISSING` nodes the edit introduced.

```
── parse ──
:42 MISSING ;
:88 ERROR }
```

## Why

An edit that drops a brace or a semicolon currently writes silently — the agent only finds out on the next read or build. This surfaces edit-introduced syntax breakage immediately, at the point of the edit.

## Design

- **Only new errors.** Pre-existing errors stay silent, even when the edit shifts their line numbers. The diff is a **non-positional multiset** keyed on `(ErrorKind, detail)` — a positional key would false-flag pre-existing errors below the edit site.
- **Advisory only.** The write always succeeds; the parse block is output, never a gate. The edit is never rejected.
- **Grammar-less languages short-circuit** to no block (returns `None`).
- Output caps at 10 listed errors, then `... and N more (X new, Y total)`.

## Scope / provenance

Slice 3 extracted from the fold-in mega-PR #112. Authored on an older base (before the `mcp.rs` module split), so a clean cherry-pick does not apply — **re-applied by hand** onto the current post-split layout:

- `src/edit_parse_check.rs` — new module (ported verbatim, incl. its test suite).
- `src/lib.rs` — module declaration.
- `src/edit.rs` — threads `parse: Option<String>` through `EditResult::Applied` and appends the block in the render path. No `mcp/tools/edit.rs` change needed — `tool_edit` forwards `apply_batch`'s output unchanged.

## Heads-up: races open PR #124

PR #124 (`feat(edit): allow tilth_edit to create new files`) also rewrites `src/edit.rs`. Intent is to **land this slice first**; #124 rebases onto it (its changes are in `apply_edits`' create path, orthogonal to the parse-check wiring in the return/render path).

## Tests

`src/edit_parse_check.rs` carries its own suite (clean edit → no block; introduced error → block; pre-existing error silent; shifted pre-existing error silent; `MISSING` reported; format cap; grammar-less → `None`; multiset surplus/key-distinctness; new-vs-total summary). Plus integration assertions in `src/edit.rs` that the rendered `apply_batch` output carries/omits the block.

Gates: `cargo fmt --check`, `cargo clippy -- -D warnings`, `cargo test` — all green.

### DIFF
--- a/src/edit.rs
+++ b/src/edit.rs
@@ -49,6 +49,9 @@ enum EditResult {
         diff: String,
         /// Hashlined context around edit sites (existing behavior).
         context: String,
+        /// Rendered `── parse ──` block when the edit introduced new
+        /// tree-sitter `ERROR`/`MISSING` nodes; `None` otherwise.
+        parse: Option<String>,
     },
     /// One or more hashes didn't match current content.
     HashMismatch(String),
@@ -67,6 +70,7 @@ fn apply_edits(path: &Path, edits: &[Edit]) -> Result<EditResult, TilthError> {
         return Ok(EditResult::Applied {
             diff: String::new(),
             context: String::new(),
+            parse: None,
         });
     }
 
@@ -264,7 +268,15 @@ fn apply_edits(path: &Path, edits: &[Edit]) -> Result<EditResult, TilthError> {
     let diff = format_diffs(&diffs);
     let context = contexts.join("\n---\n");
 
-    Ok(EditResult::Applied { diff, context })
+    let parse = crate::edit_parse_check::check(path, &content, &output)
+        .as_ref()
+        .map(crate::edit_parse_check::format_report);
+
+    Ok(EditResult::Applied {
+        diff,
+        context,
+        parse,
+    })
 }
 
 /// Format per-edit diffs as compact `-`/`+` blocks with hashline anchors.
@@ -471,7 +483,11 @@ fn render_applied(
     show_diff: bool,
 ) -> Result<String, String> {
     match apply_edits(path, edits).map_err(|e| e.to_string())? {
-        EditResult::Applied { diff, context } => {
+        EditResult::Applied {
+            diff,
+            context,
+            parse,
+        } => {
             let mut output = String::new();
             if show_diff && !diff.is_empty() {
                 output.push_str(&diff);
@@ -481,6 +497,12 @@ fn render_applied(
             }
             if !context.is_empty() {
                 output.push_str(&context);
+            }
+            if let Some(parse) = parse {
+                if !output.is_empty() {
+                    output.push_str("\n\n");
+                }
+                output.push_str(&parse);
             }
             let abs_path = fs::canonicalize(path).unwrap_or_else(|_| path.to_path_buf());
             let scope = crate::lang::package_root(&abs_path).map_or_else(
@@ -529,7 +551,7 @@ mod tests {
 
         let result = apply_edits(&path, &edits).unwrap();
         match result {
-            EditResult::Applied { diff, context } => {
+            EditResult::Applied { diff, context, .. } => {
                 assert!(
                     diff.contains("- 2:"),
                     "diff should have removed line: {diff}"
@@ -693,7 +715,7 @@ mod tests {
 
         let result = apply_edits(&path, &[]).unwrap();
         match result {
-            EditResult::Applied { diff, context } => {
+            EditResult::Applied { diff, context, .. } => {
                 assert!(diff.is_empty(), "diff should be empty for no edits");
                 assert!(context.is_empty(), "context should be empty for no edits");
             }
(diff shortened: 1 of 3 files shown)


jahala (commented):
Reviewed by Opus 4.7:

Thanks for this — the design is right (advisory check that never blocks the edit, in-memory pre/post comparison so no TOCTOU window) and gates are all green (build, 414 tests, clippy, fmt). One real correctness bug worth fixing before merge, plus one nice-to-have:

**MEDIUM — `collect_errors` double-counts MISSING via ERROR-child walk** (`src/edit_parse_check.rs:115-133`). The walker recurses into ERROR nodes' children. MISSING nodes commonly appear as synthetic children INSIDE ERROR subtrees, so they get pushed twice — once via the ERROR ancestor's recursion and once via direct visit. The multiset subtract in `diff_new_errors` can't absorb both copies against a single pre-existing entry, so an unrelated edit leaks one of them as "new." The existing `preexisting_error_silent` test happens not to catch it because its fixture (unclosed brace) doesn't produce a nested MISSING; a fixture like `if (a) ` (missing both `)` and `;`) should surface it. Fix: `return` after pushing the ERROR record; recurse only on non-error nodes.

**MEDIUM — silent parse failure indistinguishable from clean parse** (`src/edit_parse_check.rs::check`). Tree-sitter `parser.parse()` returns `None` on internal failure (OOM, timeout, allocation). The check then returns `None`, which the caller can't distinguish from "no new errors." Either log to stderr, or add an explicit comment noting that parse-failure-on-write currently appears as a silent pass — your call. Edge case, but the asymmetry between "parsed clean" and "couldn't parse" is the kind of thing that's easier to fix now than after it's wired in.

**Nits (non-blocking):**
- The `+1` capacity hint in `truncate_chars`'s truncating branch is off by 4 bytes (`…` is 3 bytes in UTF-8 plus 1, accounts for 0 rather than the ellipsis width).
- `pub` items inside the `pub(crate) mod edit_parse_check` could be `pub(crate)` — the wider visibility doesn't reach the user since the module is gated.
- The `+1` ellipsis-presence bound in the truncation test could use a one-line comment explaining what it bounds against.

Verified clean:
- Pre/post content both read from in-memory snapshots — no TOCTOU vs the write
- Each `apply_edits` creates its own tree-sitter `Parser` — no Send concerns under rayon
- `OutlineCache` + `BloomFilterCache` are mtime-keyed; the edit's write naturally invalidates on next access, no manual wiring required
- Grammar-less languages short-circuit cleanly via `outline_language()` returning None
- Advisory semantics: write always succeeds first; parse-check failure never aborts the edit
- MCP `isError: false` preserved for parse-flagged successes

paulnsorensen:
Thanks for the close read — landed four of these, pushing back on two. Tested everything directly rather than reasoning from the diff.

**Fixed:**
- **The "new error" leak you sensed is real — but the cause and fix were different.** Reproduced a false positive in Python: a pre-existing unclosed `(` makes the ERROR node swallow the line below it into its *text*, so editing that unrelated line changes the multiset key and leaks a "new" error. Fix: cap ERROR `detail` at its first line (stable across edits beneath the bracket) + regression test. Note this is the *top-level* ERROR node, not a nested child — see the first pushback.
- **Silent parse failure:** took the comment option — documented that `parser.parse() -> None` reads as a clean parse, an intentional advisory limitation. Skipped stderr logging since the path is effectively unreachable (tree-sitter is error-tolerant, language always set, no timeout configured) and I didn't want defensive code for an impossible case.
- **`pub` -> `pub(crate)`:** done across the module.
- **Test `+1` comment:** added.

**Pushing back on two:**
- **`collect_errors` double-counts MISSING via ERROR-child walk → couldn't reproduce.** The walker is single-visit (each node pushed at most once), so there's no double-count, and the `if (a)` fixture produces one ERROR record with no nested MISSING. I applied the exact `return`-after-ERROR fix and re-ran: the leak still fires (it's the top-level node), while introduced-error detail drops from 2 records to 1 — a net regression. So I fixed the real leak (above) instead.
- **`truncate_chars` `+1` off by 4 bytes → moot.** There's no `+1` in `truncate_chars` (it's `with_capacity(max)`), and the truncating branch re-allocates via `format!`, so the hint never applies to truncated output. Possibly written against an earlier revision.

Gates green: 415 tests, clippy `-D warnings`, fmt. Pushed as `71e12c5`.

---
agent on behalf of; paulnsorensen

jahala (approved):
Reviewed by Opus 4.7: pushback on my collect_errors diagnosis was correct — the multiset already tracks ERROR and MISSING under distinct keys; my proposed return would have been a regression. The real bug you reproduced (ERROR detail mutating with downstream text) is well-fixed by the first-line cap, and the new regression test pins it. Silent parse failure: the inline comment is enough. All 419 tests pass, gates clean. Approve.

PR #127, opened 2026-05-23
feat(read): budgeted sections, signature/stripped modes + default response cap

paulnsorensen:
Slice 4 of the fork fold-in (jahala/tilth#112), extracted as a standalone PR and ported onto the post-`mcp/`-split layout. Orthogonal to the open edit (#124) and grok (#125) work — touches only read/budget/types.

## What this adds

- **`tilth_read` `mode` parameter**: `auto` (default) | `full` | `signature` | `stripped`. The existing `full: true` flag becomes a legacy alias for `mode="full"`.
- **`signature` mode** — hash-prefixed declaration lines only, no bodies, for code files (non-code falls back to the normal read).
- **`stripped` mode** — whole-file content with plain comments, debug logs, and collapsible blank runs removed; plain right-aligned line numbers, no hash anchors (an explicitly non-editable view).
- **Budgeted multi-section reads** (`read_ranges_with_budget`) — when `budget` is set on a `sections` call, blocks are emitted in user order, a later over-budget section is replaced by an omission marker (while the marker fits), an over-budget first block is truncated, and the output never exceeds the budget.
- **`budget::apply` truncation bugfix** — reject a `\n\n` cut at byte 0 (code-format output's only `\n\n` is the header/body separator, so the old code truncated to an empty body) and clamp the no-boundary fallback to a UTF-8 char boundary instead of raw `max_bytes` (which could panic on emoji-dense single-line content).

## Porting notes

- Re-homed the original `src/mcp.rs` `tool_read` hunks into `src/mcp/tools/read.rs` (helpers + dispatch) and the schema change into `src/mcp/tools/definitions.rs`, since `mcp.rs` was split after this slice's base commit.
- The fork's unused `apply_with_info` / `TruncationInfo` scaffolding is intentionally omitted (YAGNI — nothing in this slice consumes the metadata); the truncation fix is folded directly into `apply`.
- No prompt changes, so `AGENTS.md` regen is a no-op and the byte-lock tests are unaffected.

## Verification

`cargo build --release`, `cargo clippy -- -D warnings`, `cargo fmt --check`, and `cargo test` all clean (416 tests pass, including 12 ported + 4 hardening tests covering the budget bugfix, budgeted multi-section invariants, mode dispatch, unknown-mode error, non-code fallbacks, and the `full`/`mode="full"` equivalence).

### DIFF
--- a/src/budget.rs
+++ b/src/budget.rs
@@ -1,5 +1,10 @@
 use crate::types::estimate_tokens;
 
+/// Default cap applied to MCP tool responses when no explicit `budget` is given.
+/// Sits just under the host's ~25K-token tool-response limit so a broad
+/// `tilth_read full:true`, `regex`, or `diff` can't blow the host budget.
+pub const DEFAULT_BUDGET: u64 = 24_000;
+
 /// Apply token budget to output. Works backwards from the cap:
 /// 1. Reserve 50 tokens for header
 /// 2. Truncate content at section boundaries to avoid broken output
--- a/src/budget.rs
+++ b/src/budget.rs
@@ -26,12 +26,22 @@ pub fn apply(output: &str, budget: u64) -> String {
     let safe_max = body.floor_char_boundary(max_bytes);
     let truncated = &body[..safe_max];
 
-    // Prefer section boundaries (\n\n##) to avoid cutting mid-match in search results
+    // Prefer section boundaries (\n\n##) to avoid cutting mid-match in search results.
+    // Fallback is `safe_max` (= truncated.len()), never `max_bytes`: `max_bytes` may
+    // land mid-UTF-8-codepoint and would panic `&body[..cut_point]` on emoji-heavy
+    // single-line content with no newline in the truncated region.
+    //
+    // Reject `\n\n` cuts at position 0: body always starts with the structural
+    // header/body separator, and for code-rendered output (every line carries
+    // a `<n>:<hash>|` prefix, so blank source lines are still non-empty) that's
+    // the *only* `\n\n` in the body. Without this filter, every truncated code
+    // file would return zero content lines.
     let cut_point = truncated
         .rfind("\n\n##")
-        .or_else(|| truncated.rfind("\n\n"))
+        .filter(|&p| p > 0)
+        .or_else(|| truncated.rfind("\n\n").filter(|&p| p > 0))
         .or_else(|| truncated.rfind('\n'))
-        .unwrap_or(max_bytes);
+        .unwrap_or(safe_max);
 
     let clean_body = &body[..cut_point];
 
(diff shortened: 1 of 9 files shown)


jahala (commented):
Reviewed by Opus 4.7:

Thanks for the careful extraction — the scope is clean and the implementation is solid. All four gates pass (build, 416 tests, clippy, fmt). Two things worth addressing before merge, one minor thing to clarify:

**MEDIUM — `mode=signature`/`stripped` silently ignored when `section` or `sections` is also set** (`src/mcp/tools/read.rs:127`). If a caller passes both, they get a plain section read with no indication the mode was dropped. An agent that asked for signatures of a range will get full content and not know why. Easiest fix is to return an error: `mode=signature/stripped is not compatible with section/sections`. Alternatively, document the precedence in the schema description — but the error is cleaner since the combination is almost certainly a caller mistake.

**MEDIUM — `cache` parameter is unused in the code path it was designed for** (`read.rs:157`, `:204`). Both `read_signature_file` and `read_stripped_file` accept `cache: &OutlineCache` but only forward it to the non-code fallback. The hot path — `get_outline_entries` — re-parses tree-sitter fresh on every call. For large files or repeated calls this is measurably slower than the cached normal-read path. Minimal fix: rename to `_cache` and add a `// TODO: wire into cache.get_or_compute once OutlineCache stores Vec<OutlineEntry>` comment so the gap is visible rather than hidden.

**Nit — `full: true` + `mode: "signature"` conflict** (`read.rs:32`): signature silently wins. Add a single comment `// signature/stripped take priority over force_full` and ideally a test case so the behavior is pinned against future refactoring.

**Other minor items:**
- `read_signature_file` builds a `ViewMode::Signature` header before the `FileType::Code` branch check, then discards it on the non-code fallback (line 180 returns early without the header). Moving the allocation inside the `Code` arm would clarify intent.
- `strip_noise` does not strip `/* … */` block comments for Go, C, C++ (only `//` line comments). Pre-existing limitation of `strip.rs`, but this PR makes it user-facing via the stripped mode — worth a comment in `strip.rs` so contributors know the scope.

The budget bugfix, the `collect_blocks` refactor, and the mode validation are all well done. The test for `full:true` alias equivalence on a file that actually exceeds `TOKEN_THRESHOLD` is particularly good — that's a real assertion, not a vacuous one.

Fork-baggage check passes cleanly: diff is confined to read/budget/types, no `canonicalize` / `path_guard` / scope-check logic, no vestigial #112 references.

paulnsorensen:
Thanks for the review — all five items addressed in `a43561c`, with a scope note at the end.

**MEDIUM — `mode=signature/stripped` silently ignored with `section`/`sections`** (`read.rs:127`)
Fixed: now returns an error (`mode=… cannot be combined with section/sections`) alongside the existing `section`+`sections` mutual-exclusion guard, so the dropped-mode case fails loud instead of returning a plain section read. Pinned by `tool_read_signature_mode_rejects_section`.

**MEDIUM — dead `cache` param on the code path** (`read.rs:157`, `:204`)
Documented rather than wired — full wiring is out of scope for this PR (`OutlineCache` stores formatted outline strings, not `Vec<OutlineEntry>`). One deviation from the suggested fix: I kept the param named `cache`, not `_cache`. Clippy's `used_underscore_binding` fires because the param is genuinely live on the non-code fallback path, so `_cache` doesn't compile here. Added a tracking comment on the code path marking the cache-wiring gap instead.

**Nit — `full:true` + `mode:signature` precedence** (`read.rs:32`)
Added the `// signature/stripped take priority over force_full` comment and a pinning test (`tool_read_signature_beats_full_flag`).

**Minor — discarded signature header on non-code fallback** (`read.rs:176`)
Moved the `file_header` allocation inside the `Code` arm so it's only built when used.

**Minor — block comments not stripped** (`strip.rs`)
Added a scope comment noting stripping is line-comment-only and that a `/* … */` stripper needs multi-line state (out of scope here). Small correction to the note: it affects all languages, not just Go/C/C++ — the comment reflects that.

**Scope note:** while this branch was open I also folded in the G2 default-response-cap fix (`333cc91`, extracted from #38). `apply_budget(.., None)` was previously uncapped, letting a broad `full:true` / regex / diff blow the host's ~25K tool-response limit; it now applies `budget::DEFAULT_BUDGET` (24_000) on the `None` arm. Happy to split that back into its own PR if you'd rather keep this slice read-only — just say the word.

Gates green: build / clippy `-D warnings` / fmt / 419 tests.

jahala (approved):
Reviewed by Opus 4.7: both MEDIUMs from the previous review cleanly addressed (mode+section now hard errors, cache parameter kept with clear tracking comment). The new DEFAULT_BUDGET cap is the right fix for unbudgeted callers — one note for follow-up: tilth_diff still bypasses apply_budget, so the diff-blows-budget motivation isn't fully covered for that tool. Not a blocker since diff has its own truncation. 434 tests pass, gates clean. Approve.

PR #130, opened 2026-05-27
feat(mcp): tilth_write — hash/overwrite/append modes (alternative to #124)

paulnsorensen:
## Proposal: unify file creation into a mode-based `tilth_write`

This is an **alternative to #124** (`feat(edit): allow tilth_edit to create new files`). Rather than add a `create: true` discriminator to `tilth_edit`, it renames the tool to `tilth_write` and gives it three per-file **modes**, with whole-file creation as a natural case of `overwrite`.

> #126 has since merged, so this is now a clean standalone slice against `main` (13 files).

### Modes

| mode | behavior |
|---|---|
| `hash` (default) | replace lines at hash anchors from `tilth_read` — the former `tilth_edit`, unchanged |
| `overwrite` | whole-file write from `content`; **create-only by default** (atomic `O_CREAT\|O_EXCL`), pass `overwrite: true` to replace an existing file |
| `append` | append `content`, creating the file (and parent dirs) if absent |

```json
{ "files": [
  { "path": "existing.rs", "edits": [{ "start": "42:a3f", "content": "..." }] },
  { "path": "new.rs",      "mode": "overwrite", "content": "fn main(){}\n" },
  { "path": "app.log",     "mode": "append",    "content": "started\n" }
]}
```

### Coverage vs #124 — strict superset

Everything #124 exposes is covered, plus more:

| #124 (`create: true`) | here |
|---|---|
| create new file | ✅ `overwrite`, create-only by default |
| atomic `O_CREAT\|O_EXCL` | ✅ |
| parent dirs auto-created | ✅ |
| symlink not followed on create (dangling → `AlreadyExists`) | ✅ |
| hashlined output for follow-up edits | ✅ (overwrite + append echo hashlines) |
| empty-content touch | ✅ |
| strict bool (`"true"` string rejected) | ✅ |
| **duplicate-path rejection** | ✅ — folded in, now runs up front across **all** modes (the one guarantee #124 had that the original fork design lacked) |

Beyond #124: replace-existing (`overwrite: true`), `append`, `O_NOFOLLOW` symlink-follow refusal on **both** the replace and append paths, a scope guard (refuse writes outside the resolved scope, **uniform across all modes**), and `diff: true` rendering.

### Review follow-ups

Addressed in the latest commit (review by jahala / Opus 4.7):

- **append echo TOCTOU** — the echoed hashlines are now computed from the pre-write snapshot plus the known `content`, never a post-write re-read, so a concurrent appender can't mis-number the anchors. Handles the no-trailing-newline merge.
- **hash-mode scope escape** — the scope guard now runs for **all** modes. `apply_batch`'s `package_root` is blast-radius-only and never enforced containment, so a `../../` hash entry previously traversed out of scope.
- **`append` O_NOFOLLOW** — closed the asymmetry; `write_append` refuses to follow symlinks on Unix, matching `overwrite`.
- **typed `TilthError::AlreadyExists`** — the create-only collision now surfaces through a typed variant alongside `NotFound` / `PermissionDenied`.
- **`h`/`w`/`a` aliases** documented in `mcp-edit.md`.
- Public docs (`README` / `ARCHITECTURE` / `CLAUDE.md`) renamed `tilth_edit` → `tilth_write`.

The `tilth_read` batch-only change (drop singular `path`) that briefly rode in this PR has been **split into its own PR — #132** to keep this one focused on the write surface.

### What's intentionally *not* here

The fork's `tilth_write` also bundles a hash-drift auto-fix (`auto_fix_locate` / relocation probe). That's a separate, orthogonal feature and is **left for its own follow-up PR** to keep this one reviewable.

### Migration

The tool is renamed `tilth_edit` → `tilth_write`. There's no released config that embeds the old name, but if you pinned the tool in a permission allowlist (e.g. `mcp__tilth__tilth_edit` in `.claude/settings.json`), update it to `mcp__tilth__tilth_write`.

### Tests / gates

- New: `tool_write` append anchor-numbering, hash-mode scope refusal, `write_append` symlink refusal — on top of the existing fs-op / dispatcher / schema tests.
- `cargo fmt --check`, `cargo clippy -- -D warnings` (CI form), `cargo test` — **461 pass, 0 fail**.
- `prompts/mcp-edit.md` + `prompts/mcp-base.md` updated; `AGENTS.md` regenerated; byte-lock tests updated.


paulnsorensen:
@jahala my thinking is that we align more with the tools the harnesses use by name now (Read/Write) and then also make creation on by default (so LLM doesn't have to think) and then pass a flag to overwrite a file if it wants to replace.

Claude will probably push back over users of the MCP, but I think it will only affect people who call the MCP out by name, which appears to just be me and one other person (who I can go and make a PR on their dotfiles for): https://github.com/search?q=mcp__tilth__tilth_edit&type=code

jahala (commented):
Reviewed by Opus 4.7:

Honestly this is the better design — `mode` enum is more extensible than my `create: true` flag, `append` is a genuinely useful addition, and the `O_NOFOLLOW` on the overwrite path is a security win #124 doesn't have. Happy to close #124 once this lands.

Two real blockers I want to flag before approving though:

**HIGH — `append` echo has a TOCTOU** (`src/mcp/tools/write.rs:296-303`): the success path re-reads the file after `write_append` to compute the echo lines. If anything else appends to that file between the write and the re-read, the re-read picks up more lines than this call wrote — `start_idx = total.saturating_sub(appended)` ends up wrong and the returned hashlines point to the wrong line numbers. The agent then tries a hash-mode edit against those anchors and gets a confusing hash mismatch. Cleanest fix: capture the pre-append line count *before* calling `write_append`, then slice the tail from `pre_count..` without re-reading.

**HIGH — scope guard skipped on hash mode** (`write.rs:213`): `path_within_scope` only runs for `overwrite | w | append | a`. The comment says hash mode is fine because `apply_batch` "roots paths to the package root itself" — but `apply_batch` resolves relative to cwd without doing actual containment enforcement. A `hash` entry with `../../etc/passwd` traverses out of scope. Easiest fix is just running the guard uniformly across all three modes, or pass the scope into `apply_batch` and have it enforce containment there.

**MEDIUM — a few public doc spots still say `tilth_edit`**: `README.md:225,232`, `ARCHITECTURE.md:191,736`, `CLAUDE.md:51`. README especially is the first thing a new user sees.

**Couple of bits from #124 worth pulling in while you're at it**:

1. **`TilthError::AlreadyExists` typed variant** (in `src/error.rs` on #124's branch) — right now the existing-file case in overwrite-create-only mode surfaces as raw `io::ErrorKind::AlreadyExists` text. Adding the typed variant makes it sit alongside the existing `NotFound` / `PermissionDenied` and gives MCP callers a clean class to branch on. Around 5 lines net.

2. **The dangling-symlink test** (#124's `create_file_fails_when_path_is_dangling_symlink`, `#[cfg(unix)]`). Pins the invariant that `O_CREAT|O_EXCL` doesn't follow symlinks on Linux/macOS — worth having for overwrite create-only mode the same way.

Smaller stuff (non-blocking):
- The `h`/`w`/`a` shorthand aliases aren't documented in `mcp-edit.md` — either document them or drop them, agents won't discover them otherwise.
- `append` doesn't have an `O_NOFOLLOW` equivalent (asymmetric with overwrite) — worth a code comment acknowledging the gap.
- CHANGELOG note for the rename would help integrators with existing `tilth_edit` configs know what to update.

Fix the two HIGHs, port in the typed error + symlink test, and we're good to go — I'll approve and close #124.

paulnsorensen:
Thanks for the thorough review — all of it is addressed in the latest push (`0f92e86`).

**HIGH — append echo TOCTOU** (`tools/write.rs`): fixed. The echo is now computed from the pre-write snapshot plus the known `content` (`before + content`), never from a post-write re-read, so a concurrent appender can't inflate the count and mis-number the anchors. Also handles the no-trailing-newline merge. Test: `append_echoes_appended_region_with_correct_line_numbers`.

**HIGH — scope guard skipped on hash mode** (`tools/write.rs`): fixed. The guard now runs uniformly across all modes before dispatch. You're right that the old comment was wrong — I traced it: `apply_batch`'s `package_root` (`edit.rs:507-512`) is used only for blast-radius reporting and never enforces containment, so a `../../` hash entry did traverse out of scope. Test: `refuses_hash_write_outside_scope`.

**MEDIUM — docs still say `tilth_edit`**: renamed across `README` / `ARCHITECTURE` / `CLAUDE.md` (and the README edit example, which was also still using the pre-`files[]` shape).

**Port — `TilthError::AlreadyExists`**: added the typed variant + `exit_code` arm; the create-only collision now surfaces through its `Display`.

**Port — dangling-symlink create test**: already covered by `write_overwrite_create_only_refuses_dangling_symlink` (`mcp/write.rs`), which pins exactly that `O_CREAT|O_EXCL` invariant — so I didn't add a duplicate.

**Low — `h`/`w`/`a` aliases**: documented in `mcp-edit.md` (`Mode aliases: h=hash, w=overwrite, a=append.`).

**Low — `append` O_NOFOLLOW asymmetry**: closed rather than commented — `write_append` now passes `O_NOFOLLOW` on Unix (ELOOP → InvalidInput), symmetric with overwrite. Test: `write_append_refuses_symlinks`.

**Low — CHANGELOG note**: there's no CHANGELOG in the repo, and the real breakage is `settings.json` allowlists pinning `mcp__tilth__tilth_edit` — captured as a Migration note in the PR body instead.

One thing beyond your review: I also made `tilth_read` batch-only (dropped the singular `path`; `paths` is now required) to match the running fork's contract. That's a `tilth_read` change riding in this PR — happy to split it into its own PR if you'd rather keep this one purely about `tilth_write`.

Gates: 438 tests pass, `clippy -- -D warnings` clean, `fmt --check` clean.

---
agent on behalf of; paulnsorensen

paulnsorensen:
Heads-up: `main` advanced while I was working (it picked up #127's `signature`/`stripped`/`mode` read views), so I merged it in. The `tilth_read` paths-only change reconciles cleanly with that work — they compose:

- `mode` (and its legacy `full` alias) reshape per file, so they're allowed in batch reads.
- `section` / `sections` stay single-file only.
- `paths` is required; singular `path` is gone.

Updated #127's read-mode tests to the `paths: [...]` form. CI is green (fmt / clippy `-D warnings` / 462 tests), PR is mergeable.

---
agent on behalf of; paulnsorensen

paulnsorensen:
Split the `tilth_read` paths-only change out into #132 so this PR stays focused on the write surface — `tilth_read` here is now unchanged from main. Rebased on latest main; gates green (461 tests, clippy `-D warnings`, fmt).

---
agent on behalf of; paulnsorensen

PR #132, opened 2026-05-30
feat(mcp): make tilth_read batch-only (paths required, drop singular path)

paulnsorensen:
Makes `tilth_read` batch-only: a required `paths` array replaces the singular `path` parameter. Pass a single-element array to read one file.

### Why
The schema accepted both `path` and `paths`, which left agents guessing which one a call wanted. The running fork has been batch-only for a while; this brings upstream in line.

### Changes
- Schema: `required: ["paths"]`, drop `path`, bound `paths` to 1–20.
- `section` / `sections` stay single-file controls — rejected when more than one path is given.
- `mode` (and its legacy `full` alias) reshape per file, so they remain valid in batch reads.
- Prompt (`mcp-base.md`) + `AGENTS.md` updated to the batch-only wording; `SERVER_INSTRUCTIONS` byte-lock refreshed.

Composes cleanly with the `signature` / `stripped` read views from #127.

### Migration
`tilth_read` callers must now pass `paths: [...]` (an array) — a singular `path` is no longer accepted.

### Gates
`cargo fmt --check`, `cargo clippy -- -D warnings`, `cargo test` — **440 pass, 0 fail**.

> Split out of #130 (tilth_write modes) to keep that PR focused on the write surface.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -26,9 +26,9 @@ Re-expanding a previously shown definition returns [shown earlier].
 tilth_read: Read file content with smart outlining. Replaces cat/head/tail.
 Batch-only: ALWAYS pass paths: [...] as an array, even for one file. DO NOT use a singular `path`.
 Small files → full content. Large files → structural outline.
-For one file you may also pass section ("<start>-<end>" or "<heading text>"), sections (array of ranges), or full.
+For one file you may also pass sections (array of "<start>-<end>" or "<heading text>" ranges; single-element array for one range) or full.
 Output:
-<line_number> │ <content>                  ← full/section mode
+<line_number> │ <content>                  ← full/sections mode
 [<start>-<end>]  <symbol name>             ← outline mode
 
 tilth_files: Find files by glob pattern. Replaces find, ls, pwd, and the host Glob tool.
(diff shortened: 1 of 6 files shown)


jahala (commented):
Reviewed by Opus 4.8:

One direction check before this lands, since it's a breaking change to the most-used tool.

The reasoning — drop the `path`/`paths` ambiguity so weak models can't trip over which to use — is sound, and we buy it's a real foot-gun. What we'd want before merging a *breaking* change to `tilth_read` is evidence it moves the needle: a small haiku run on the hard tasks (`rg_search_dispatch`, `gin_servehttp_flow`) showing tool-call success / cost-per-correct holds or improves with paths-only vs the dual API. Neutral-or-better → ship; if weak models do *worse* wrapping every single read in an array, that changes the calculus.

Got any before/after numbers from the fork, or a sense of how haiku handled array-only in practice? If not, happy to run it on our side before merge — just want the breaking change earning its keep rather than aesthetic.

jahala:
Closing as superseded-in-approach, not rejected-in-diagnosis. The problem this PR attacks is confirmed real by two independent measurements since it was opened: agents batch reads only ~6-8% of the time, in benchmarks and in organic sessions alike, and a shipped instruction telling them to batch moved nothing. Your instinct that the schema itself has to do the work is a live hypothesis.

We're currently testing the non-breaking version of that idea first — in-response coalescing hints that arrive at the moment of need — because it risks nothing for existing sessions. If its A/B fails on adoption, schema-level enforcement (this PR's mechanism) is the explicitly-acknowledged plan B and would return as a fresh experiment on a current branch rather than reviving this one against a moved main. Either way the batching problem is being actively worked, and this PR called it early.

PR #136, opened 2026-05-30
test(benchmark): add multi-edit, render-cascade, and route-logic tasks

paulnsorensen:
## What

Adds net-new benchmark task modules carved out of the fork:

- **multi-edit tasks** (`fastapi_multi_edit`, `gin_multi_edit`) — tasks requiring multiple coordinated edits
- **gin render cascade / runtime** and **route-logic** tasks
- **grok tasks** — "understand this symbol" tasks targeting `tilth_grok`
- a **no-git task variant** (`hide_git`): moves `.git` aside for the agent run so the bug must be localized by reading/navigating code, not `git blame`/`log`/`diff`

## Wiring

- `Task.apply_mutations` gains a `hide_git` branch (default `False` — existing behavior unchanged)
- `fixtures/reset.py` restores `.git` before each reset (idempotent + crash-safe, so a timed-out no-git run recovers on the next task)
- new tasks registered in `tasks/__init__.py`

## Notes

All net-new files plus three small, additive wiring edits — zero conflict with `origin/main`. The new fastapi task uses the same `/usr/local/bin/python3.10 -m pytest` invocation as the existing fastapi tasks (no new env dependency). Excluded from this PR: unrelated drive-by edits in the fork to `run.py`/`analyze.py`/`config.py` and a `uv`-migration of existing tasks. `python3 -c "import tasks"` registers all 51 tasks cleanly.

### DIFF
--- a/benchmark/fixtures/reset.py
+++ b/benchmark/fixtures/reset.py
@@ -15,6 +15,14 @@ def ensure_repo_clean(repo_path: Path, pinned_sha: str | None = None) -> None:
     Handles both uncommitted changes (dirty working tree) and committed
     mutations (extra commits on top of the pinned SHA).
     """
+    # Restore `.git` if a no-git task (hide_git) moved it aside. Idempotent and
+    # crash-safe: runs before every reset, so a timed-out/crashed no-git run is
+    # recovered on the next task's pre-reset.
+    git_dir = Path(repo_path) / ".git"
+    hidden = Path(repo_path) / ".git_hidden"
+    if hidden.exists() and not git_dir.exists():
+        hidden.rename(git_dir)
+
     needs_reset = False
 
     # Check for dirty working tree
--- a/benchmark/tasks/__init__.py
+++ b/benchmark/tasks/__init__.py
@@ -28,6 +28,7 @@ from .fastapi_edit_tasks import (
     FastAPIEditResponseFilterTask,
     FastAPIEditScopeCacheTask,
 )
+from .fastapi_multi_edit_tasks import FastAPIMultiResponseTask
 from .gin_tasks import (
     GinRadixTreeTask,
     GinClientIPTask,
@@ -40,6 +41,11 @@ from .gin_edit_tasks import (
     GinEditAbortCheckTask,
     GinEditContextResetTask,
 )
+from .gin_multi_edit_tasks import GinMultiContextTask
+from .gin_render_cascade_tasks import GinRenderContractCascadeTask
+from .gin_render_runtime_tasks import GinRenderRuntimeCascadeTask
+from .gin_route_logic_tasks import GinRouteCatchAllLogicTask
+from .gin_route_logic_nogit_tasks import GinRouteCatchAllNoGitTask
 from .express_tasks import (
     ExpressJsonSendTask,
     ExpressRenderChainTask,
@@ -80,6 +86,7 @@ TASKS = {
     "rg_edit_line_locate": RipgrepEditLineLocateTask(),
     "rg_edit_preceding": RipgrepEditPrecedingLinesTask(),
     # fastapi (Python)
+    "fastapi_edit_multi_response": FastAPIMultiResponseTask(),
     "fastapi_dependency_resolution": FastAPIDependencyResolutionTask(),
     "fastapi_request_validation": FastAPIRequestValidationTask(),
     "fastapi_depends_internals": FastAPIDependsInternalsTask(),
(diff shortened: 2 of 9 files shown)


jahala (commented):
Reviewed by Opus 4.8:

The render-cascade, render-runtime, and route-logic tasks are well-designed — novel mutations, test-command validation (no brittle required-strings), and `hide_git` is a clever control for the "no git-blame oracle" condition. A few things:

**[fix] `gin_edit_multi_context` recycles all three mutations from existing single-bug tasks.** The three mutations (`c.index++ → += 2`, `c.index >= abortIndex → >`, `c.index = -1 → 0`) are byte-identical to `gin_edit_tasks.py:25, 69-70, 113-114`. A multi-edit task should measure multi-bug *localization*; with three already-known bugs in the same file it measures recall of prior task solutions instead — an agent that ran the single-bug tasks pattern-matches all three sites immediately. Swap in at least one genuinely novel bug so the task actually exercises simultaneous multi-site discovery.

**[note] `.git` stays hidden through `check_correctness` for `hide_git` tasks** (restored only at the next task's reset). The inherited `go test` command doesn't need git, so it works today — but it's a latent footgun for any future `hide_git` task whose test command uses `git diff`. Consider restoring `.git` before correctness-checking rather than relying on the next reset.

**[nit]** Title says "and grok tasks" but those already merged separately — none are in this diff. Worth dropping from the title so the PR history reads cleanly.

Test-only and additive, so low risk overall — but the recycled-mutation task would produce a misleading multi-edit signal, so worth fixing before it feeds a benchmark run.

paulnsorensen:
Thanks for the review — all three addressed in 88d6dc6 + 7e1d09c.

**[fix] recycled mutations** — confirmed: `gin_edit_multi_context` reused the exact three single-bug mutations, so it measured recall of prior solutions, not multi-site localization. Swapped the recycled abort-check (`c.index >= abortIndex` → `>`) for a novel `Copy()` regression (`cp.index = abortIndex` → `cp.index = 0`, caught by `TestContextCopy`), and dropped the method/bug naming from the prompt so the agent has to localize all three from the failing tests. Verified end-to-end: all three target tests (`TestMiddlewareGeneralCase`, `TestContextCopy`, `TestContextReset`) fail with the mutations applied and pass once reverted.

**[note] `.git` through `check_correctness`** — good catch. Now restoring `.git` at the top of `check_correctness` rather than relying on the next reset, so a future `hide_git` task whose test command shells out to git can't run against the agent-run state.

**[nit] title** — dropped "and grok tasks" (those merged separately).

---

jahala:
Review by Opus 4.8:

Verified all three of your fixes and pushed one follow-up (`891e617`, maintainer edit on your branch).

- **`.git` restore** ✓ — `check_correctness` and `ensure_repo_clean` both restore it now, guarded so non-`hide_git` tasks are untouched. No regression risk to existing tasks.
- **title** ✓ — "and grok tasks" dropped; no grok files in the diff.
- **recycled mutations** — your `Copy()` swap is genuinely novel, but two of the three were still byte-identical to single-bug tasks: `c.index++ → c.index += 2` is `gin_edit_middleware_skip` (gin_edit_tasks.py:24-25) and `c.index = -1 → c.index = 0` is `gin_edit_context_reset` (113-114). So the recall confound was reduced, not eliminated. I swapped the `Next()` bug for a novel handler-loop bound flip (`c.index < safeInt8(len(c.handlers))` → `>`), so **2 of 3 sites are now novel**. `reset()` necessarily mirrors `gin_edit_context_reset` since that's the only index mutation `TestContextReset` gates — documented in the docstring.

  Verified end-to-end against gin @ `d7776de`: with all three mutations applied, `TestMiddlewareGeneralCase` → `""`, `TestContextCopy` → `0`, `TestContextReset` → `0` (all fail); the clean baseline passes all three. Each `original` string matches uniquely.

One still-open item: `fastapi_edit_multi_response` wasn't in the first review, and I can't run it here — its `/usr/local/bin/python3.10` test command (the repo-wide convention) isn't on this box. Its three mutations rely on first-occurrence `replace`, and `    if field:` is a generic substring; can you confirm the three named tests actually fail→pass with the mutations applied, the way you verified the gin task? Once that's signed off the multi-context task is ready for a scored run. Not merging.

jahala:
Independent verification of the open item (`fastapi_edit_multi_response`) — **signed off, merging**.

Ran under a uv-provisioned Python 3.10.20 against the pinned fastapi commit: pristine tree → all 3 named tests pass; all 3 mutations applied → suite fails; revert → passes again (clean reset, mirrors the harness).

Per-mutation isolation (each applied alone):
| mutation | caught by |
|---|---|
| `if field:` → `if not field:` | test_filter_second_level_model + test_return_exclude_none |
| `by_alias` flip | test_read_dict_by_alias |
| `exclude_none` flip | test_return_exclude_none |

One subtlety found and resolved: with **all three** mutations live, the `by_alias` test passes coincidentally (the gate mutation routes dict responses through `jsonable_encoder`, whose defaults happen to match that test's expectation). This does **not** weaken the task: the masking only exists while the gate bug is unfixed — fixing the gate makes the alias mutation reachable again and its test fails, so the suite passes only when all three regressions are fixed. Verified empirically, not just by inspection.

The hardcoded `python3.10` interpreter path matches the existing edit-task convention in this repo, so no change requested. Good tasks — the `hide_git` control in particular fills a measured gap (it isolates git-history-oracle solutions from real navigation).

PR #137, opened 2026-05-30
feat(mcp): conditional reads/searches via if_modified_since

paulnsorensen:
## What

Adds an opt-in `if_modified_since` ISO-8601 timestamp parameter to **`tilth_read`** and **`tilth_search`**.

When supplied:
- **read** — files whose mtime is `<=` the timestamp collapse to a one-line `# <path> (unchanged @ <ts>)` stub instead of their content (batch and single).
- **search** — result *sections* for unchanged files collapse to the same stub.
- both tools prepend a `{"if_modified_since":"<now>"}` token header so the caller can capture a fresh timestamp to echo on the next call.

## Why

Lets an agent re-read or re-search without paying for content it already has — the server replies "unchanged" instead of re-sending the body. The round-trip token means the agent never has to manufacture its own timestamp.

## Design notes

- **Strictly opt-in.** With the parameter absent, output is byte-identical to today — no header, no stubbing. All 447 pre-existing tests stay green; 6 new tests cover the stub/content/header behavior on both tools plus the schema.
- A **missing file is treated as changed** (rendered as-is, never stubbed) — `file_changed_since` returns `true` on a metadata error.
- Timestamp conversion is a new self-contained **`src/mcp/iso.rs`** using Howard Hinnant's proleptic-Gregorian date algorithms — **no `chrono` dependency**.

## Scope

Carved out of the fork (`paulnsorensen/tilth`). This is the "net-new `iso.rs` + handler wiring" slice. The fork's own version is entangled with two unrelated change-sets (a path-suffix grammar and a budget-truncation-metadata header); this PR deliberately ports **only** the `if_modified_since` feature, fitted to upstream's current read/search handlers. Builds clean; `fmt --check`, `clippy -D warnings`, and the full test suite all pass.

### DIFF
--- a/src/mcp/iso.rs
+++ b/src/mcp/iso.rs
@@ -1,0 +1,222 @@
+//! ISO-8601-ish UTC timestamp helpers for the JSON cache-token header and
+//! `if_modified_since` handling. Avoids pulling chrono — uses Howard
+//! Hinnant's date algorithms for proleptic Gregorian conversion.
+
+use std::fs;
+use std::path::Path;
+use std::time::{SystemTime, UNIX_EPOCH};
+
+/// Render the bare `YYYY-MM-DDTHH:MM:SSZ` string. Callers wrap this into
+/// the JSON cache-token line or whatever surface they need.
+pub fn iso_ts(ts: SystemTime) -> String {
+    let secs = ts.duration_since(UNIX_EPOCH).map_or(0, |d| d.as_secs());
+    format_iso_utc(secs)
+}
+
+/// Parse an RFC-3339-ish timestamp `YYYY-MM-DDTHH:MM:SSZ` back to `SystemTime`.
+/// Returns None when malformed.
+pub fn parse_iso_utc(s: &str) -> Option<SystemTime> {
+    let s = s.trim();
+    let (date, time) = s.split_once('T')?;
+    let mut dparts = date.split('-');
+    let y: i64 = dparts.next()?.parse().ok()?;
+    let mo: u32 = dparts.next()?.parse().ok()?;
+    let d: u32 = dparts.next()?.parse().ok()?;
+    let time = time.trim_end_matches('Z');
+    let mut tparts = time.split(':');
+    let hh: u32 = tparts.next()?.parse().ok()?;
+    let mm: u32 = tparts.next()?.parse().ok()?;
+    let ss: u32 = tparts.next().unwrap_or("0").parse().ok()?;
+    let secs = days_from_civil(y, mo, d) * 86_400
+        + i64::from(hh) * 3600
+        + i64::from(mm) * 60
+        + i64::from(ss);
+    if secs < 0 {
+        return None;
+    }
+    Some(SystemTime::UNIX_EPOCH + std::time::Duration::from_secs(secs as u64))
+}
+
+// Howard Hinnant's days_from_civil for proleptic Gregorian dates.
+#[allow(clippy::cast_possible_wrap)]
+fn days_from_civil(y: i64, m: u32, d: u32) -> i64 {
+    let y = if m <= 2 { y - 1 } else { y };
+    let era = if y >= 0 { y } else { y - 399 } / 400;
+    let yoe = (y - era * 400) as u64;
+    let m = i64::from(m);
+    let d = i64::from(d);
+    let doy = ((153 * (if m > 2 { m - 3 } else { m + 9 }) + 2) / 5 + d - 1) as u64;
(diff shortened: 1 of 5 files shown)


jahala (commented):
Reviewed by Opus 4.8:

Clean extraction, and the `no_chrono` ISO parser is a nice constraint to hold. The strictly-opt-in design is sound and the happy-path tests are solid. Two things to fix before merge — both verified against the branch:

**[HIGH] Equal-mtime → stale content on coarse-granularity filesystems.** `file_changed_since` uses strict `mtime > since` (`src/mcp/iso.rs:116`), and the token echoed to the caller is `SystemTime::now()` captured at handler entry (`src/mcp/tools/read.rs:284`, before any file is read). On filesystems with 1s mtime resolution (macOS HFS+, NFS, ext3/relatime), a file written *after* the read but within the same second lands with its mtime truncated below the full-precision `since` → `mtime > since` is false → the file is stubbed as "unchanged" on the next call and the agent silently acts on stale content. Two complementary fixes:
1. `mtime >= since` in `file_changed_since` (equal mtime = "might have been written this instant, treat as changed").
2. Capture `now` *after* the reads complete, right before `with_meta_header` — not at handler entry. A slow multi-file batch widens the window.

**[MED] `parse_iso_utc` silently accepts out-of-range dates.** There's no calendar-range guard (`iso.rs:29` parses `mo`/`d`/`hh`/… as bare `u32`), and `parse_iso_malformed_returns_none` is vacuously true (`...is_some() || ...is_none()`, `iso.rs:142-144`). A client that sends `2026-13-99T…` gets a plausible-but-wrong `SystemTime` and makes silently-wrong caching decisions. Add range guards (`mo: 1..=12`, etc.) and assert `is_none()` for an out-of-range input.

Also worth a test pinning the equal-mtime boundary — the batch test sidesteps it with ±1 day. Everything else is in good shape.

jahala (commented):
Reviewed by Opus 4.8:

A direction question alongside the correctness notes, since you'll have thought about this more than us.

`if_modified_since` is squarely on-mission (cut re-sent tokens), but it overlaps with the **session dedup** tilth already does — the server tracks what it expanded and returns `[shown earlier]` for repeats. The difference as we see it: dedup is automatic + server-side per session; `if_modified_since` is explicit + client-driven and survives across whole-file reads / tool boundaries / process restarts. Complementary, but real overlap.

What we're chewing on: does the explicit version earn its API surface (a timestamp param + the mtime-comparison correctness surface) on top of the automatic dedup that's already there? Put differently — what does `if_modified_since` buy in your workflows that session dedup doesn't?

Not a blocker — just want to be adding a distinct capability rather than a second mechanism for the same goal. How do you see the two relating (or should dedup just be extended instead)?

jahala:
Maintainer decision: HOLD — not merging in its current shape, with reasons and a door left open.

1. **Demand evidence points the wrong way.** We measured repeat-consumption demand on real transcripts this month (repeat-question study across the benchmark corpus) and it came back thin — the same evidence that shelved our own memoization design. `if_modified_since` serves that same demand shape, and tilth already covers the in-session case with dedup (`[shown earlier]`).
2. **Permanent cost for opt-in value:** the two schema params + descriptions are paid in tokens by every session forever, while nothing in the shipped instructions would steer an agent to round-trip the token — realistically it's dead surface until proven otherwise.
3. The known mtime-race (strict `>` + token captured before I/O on 1s-resolution filesystems) and the vacuous malformed-parse test would both need fixing regardless.

What would reopen this: evidence of real re-read demand (e.g. from long-session telemetry), or a rework as an extension of the existing session dedup rather than a parallel client-driven protocol. The ISO-parsing module itself is nicely dependency-free work — worth keeping in the toolbox if a real consumer appears.

jahala:
Closing — and owning a correction in your favor first. Our earlier hold cited thin repeat-read demand from benchmark data; we've since measured real organic sessions and your premise holds there: unchanged-content re-reads run at ~31% of all reads, overwhelmingly mid-context. The demand this PR aimed at is real and our benchmark corpus couldn't see it.

Why close anyway: the same study measured how deny-shaped mechanisms behave in production — tilth's own shipped session dedup gets re-fetched by the agent within a few calls ~32% of the times it fires. Withholding content an agent turns out to need converts one saved payload into extra calls plus a fatter fetch, and a token-round-trip stub inherits that failure mode. So the successor work reworks unchanged-content handling as a whole (recovery-carrying stubs, recency bounds, A/B-gated) with this PR and those measurements as direct inputs — it'll be referenced when that lands. The dependency-free ISO-8601 module remains a nice piece; if the rework needs a timestamp channel, it's the first thing we'd reach for.

PR #138, opened 2026-05-30
feat(mcp): multi-target caller search

paulnsorensen:
## What

`tilth_search` with `kind=callers` now accepts a comma-separated query (2–5 targets) and finds the callers of each in a single batched walk, rendering one labeled section per target.

## Why

A `callers` query like `"alpha,beta"` previously searched for a single literal symbol named `alpha,beta` — which never exists — and returned an empty no-callers message. The `symbol` arm already comma-splits; the `callers` arm did not. This brings them to parity.

## How

- New `search_callers_multi_expanded` in `src/search/callers.rs` — runs one `find_callers_batch` over the deduped target set, buckets matches by target, and renders each section in caller-supplied order. Reuses the existing single-target primitives (`find_callers_batch`, `rank_callers`, `no_callers_message`, `target_seen_in_scope`).
- The `callers` arm in `tool_search` now splits the query (trim, drop empties, max 5) and dispatches: 1 target → existing `search_callers_expanded`, 2–5 → the new multi fn, >5 → an error mirroring the `symbol` arm.
- Targets are deduped first-seen so `"alpha,alpha"` collapses to a single section rather than rendering a false empty no-callers block on the second occurrence.

## Tests

Two regression tests added to `src/mcp/tools/search.rs`:
- `callers_comma_query_finds_both_targets` — both targets reported with real call sites; the literal `"alpha,beta"` is never searched as one symbol.
- `callers_duplicate_target_does_not_render_empty_section` — duplicate target collapses to one section, no false no-callers message.

Gates: `cargo build --release`, `cargo clippy -- -D warnings`, `cargo fmt --check`, `cargo test --release` (445 pass) all green.

Output is byte-identical for the existing single-target path; the multi path only activates on a comma query.


jahala (commented):
Reviewed by Opus 4.8:

The core mechanics are clean — one combined `find_callers_batch` walk, first-seen dedup, per-target buckets in caller order, the repeated-target guard. A few things before merge (verified against the branch):

**[HIGH] 2nd-hop impact analysis is silently dropped in the multi path.** `search_callers_expanded` renders an "── Adaptive 2nd-hop impact analysis ──" block (`src/search/callers.rs:369-386`, gated by `IMPACT_FANOUT_THRESHOLD`); `search_callers_multi_expanded` has no equivalent. So `"alpha,beta"` returns strictly *less* than querying each target separately — a capability regression an agent won't notice. Either port the 2nd-hop block per target (same threshold) or document the omission with a comment + a line in the PR body.

**[MED] `BATCH_EARLY_QUIT` (50) isn't scaled by target count.** The walk aborts at 50 raw hits shared across all N targets, so a popular first target can starve targets found later in the walk order. Single-target gets all 50 to itself. Suggest `batch_quit.saturating_mul(ordered.len()).min(FULL_BATCH_EARLY_QUIT)`.

**[MED] Output header diverges from single-target.** Single: `# Callers of "foo" in <scope> — N call site(s)` (H1, scope, title-case). Multi: `## callers of "foo"` + count on the next line (H2, no scope, lowercase). An agent or UI that renders differently based solely on comma-presence sees inconsistent shapes. Match the single-target header per section, or wrap the multi sections in one H1.

**[LOW]** Schema (`src/mcp/tools/definitions.rs:40`) and `prompts/mcp-base.md` still describe callers as singular — neither mentions the comma path, so an agent reading the schema won't know it exists. If you update the prompt, remember the byte-lock in `src/mcp/mod.rs` + regenerate `AGENTS.md`.

jahala:
Evidence packet for the merge. Relevance, reproduced live on main before the fix:

```
$ tilth "Trace,Apply" --callers --scope <kong checkout>
# Callers of "Trace,Apply" — no call sites found
The name "Trace,Apply" does not appear anywhere in scope.
$ tilth "Trace" --callers --scope <kong checkout>
# Callers of "Trace" — 4 call sites
```
A confident false negative — and our shipped instructions actively teach agents to pass comma-separated symbols, so exposure was growing, not shrinking.

Maintainer-side completion of the review findings (author unresponsive since 2026-06-05): the 2nd-hop impact block is now rendered per target bucket by both single- and multi-target paths (extracted, not reimplemented — and re-based on main's evolved version so the overflow line and corrected affected-count formula survive verbatim); the early-quit budget scales by deduped target count; headers render identically for single and multi. Three failing-first tests pin the behavior. Full diff-audit against main confirms nothing from main's own evolution was lost in the freshen-merge. Thanks for the port — with the parity fixes it's strictly better than N separate calls now.

PR #139, opened 2026-05-30
feat(read): fuzzy path resolution for slightly-off paths

paulnsorensen:
## What

When tilth is given a path that doesn't resolve — wrong directory component (`src/serch/foo.rs`), missing prefix (`_mcp_core.py` for `src/milknado/_mcp_core.py`), basename-only (`symbol.rs`), or a deletion typo — it now resolves to the best-matching real file and auto-opens it with an explicit correction header, instead of only suggesting a "did you mean".

## Why

Today a missing path yields a basename-only, same-directory Levenshtein suggestion (`suggest_similar` reads only `path.parent()`), so a wrong *directory* component yields nothing and the agent pays an extra round-trip. LLM path errors are structural (wrong dir / partial / plausible-nonexistent), not character-level — a subsequence/path-aware matcher is the right primitive.

## How

New module `src/read/fuzzy_path.rs` (cold-path only — never invoked on a successful read or search):

- Walks `search::walker(scope, None)` (gitignore-aware), scores every scope-relative path against the query with [`nucleo-matcher`](https://crates.io/crates/nucleo-matcher) in `match_paths` mode, and applies a confidence gate.
- **Resolved** (unique candidate, or clears `MIN_SCORE` and beats the runner-up by `MARGIN`) → caller auto-opens. **Suggestions** (ambiguous / low-confidence) → ranked "did you mean". **None** (not a subsequence of any file) → unchanged `NotFound`.
- nucleo returns `None` unless the query is a subsequence of the candidate — that hard filter rejects stale/garbage/nonexistent paths so they never auto-resolve.
- All gate thresholds (`READ_*`, `SEARCH_*`, `SUGGESTION_K`, `MAX_FUZZY_CANDIDATES`) live in one `const` block; the candidate walk is bounded and logs truncation rather than silently capping.

Wired into the three `NotFound` sites:

| Site | Profile | Behaviour |
|---|---|---|
| `read::read_file` (via new `read_file_resolving` wrapper) | `Read` (lenient) | `# <real-path> (corrected from "<query>")` header, then the file body |
| `single_query_search`, `multi_word_concept_search` (via `fuzzy_path_fallback`) | `Search` (stricter; requires a path-like query) | `# <real-path> (resolved from path-like query "<query>") · no search matches; closest file auto-opened` |

Consumers updated: the MCP single-read (`tool_read`, scope `.`) and the lib `FilePath` read both go through `read_file_resolving`.

## Scope

This implements the three core/lib `NotFound` sites the feature targets. The `Search` gate deliberately refuses to auto-open a non-path-like query (no separator/extension) even when a fuzzy candidate exists — it suggests instead, so a normal empty symbol search never surprises the caller with a file body.

Auto-opening on the *MCP merged `tilth_search`* miss is intentionally **not** included: this repo's `tool_search` dispatches per-`kind` (symbol/content/regex/callers) with no merged "any" path to hook, and a search call returning a file body is surprising enough to warrant its own change once the read-side lands. Happy to add it as a follow-up if wanted.

## Tests

`src/read/fuzzy_path.rs` ships 9 unit tests covering every acceptance case: wrong-dir/missing-prefix resolve, partial-basename resolve, deletion-typo resolve, ambiguous-basename → Suggestions (both profiles), suggestion cap at `k`, garbage path → None, `Search` refuses non-path-like, `Search` resolves a path-like single winner.

Gates: `cargo build --release`, `cargo clippy -- -D warnings`, `cargo fmt --check`, `cargo test --release` (452 pass) all green. No regression: an exact existing-path read does no walk and is byte-identical to today.


jahala (commented):
Reviewed by Opus 4.8:

Nice structure — the gate-profile split (lenient Read vs stricter Search) is sensible, and the subsequence hard-filter elegantly rejects garbage queries. One HIGH I'd want fixed before merge, plus a couple of MEDs (all verified against the branch):

**[HIGH] A unique-but-low-score candidate silently resolves to the wrong file.** In `apply_gate` (`src/read/fuzzy_path.rs:309-313`), `unique = scored.len() == 1` bypasses *both* the score floor and the margin: `clears_floor = unique || *top_score >= min_score`. So if exactly one file is a subsequence of the (wrong) query, it auto-opens regardless of score. Concretely: an agent asks for `config.ts` after it's been renamed; only `config.test.ts` remains; `config.ts` is a subsequence of `config.test.ts`; it's the unique match → its content is returned *as if it were* `config.ts`, with no confidence gate. "Unique among the scored set" ≠ "confident match" — especially after the 20k-candidate truncation. Fix: require the floor unconditionally (`clears_floor = *top_score >= min_score`) and keep `unique` only as the margin relaxation.

**[MED] Score thresholds have no empirical test backing.** `READ_MIN_SCORE=40` / `SEARCH_MIN_SCORE=80` are justified only in comments. If nucleo's scoring weights shift across a version bump, the gate degrades silently with no test failing. Add score-assertion tests (a canonical correct match clears the floor; a known incidental subsequence does not) and pin nucleo to a patch version (`= "0.3.1"`) rather than the `"0.3"` range.

**[MED] Only the single-file auto/full path gets fuzzy resolution.** The batch `paths` loop (`src/mcp/tools/read.rs:79`) calls plain `read_file`, and `signature`/`stripped` modes (`read.rs:75/77/141/145`) call their helpers directly — all bypass `read_file_resolving`. So the same slightly-wrong path resolves in single-auto mode but hard-errors in batch or signature mode. Route those through `read_file_resolving` (or catch `NotFound` and retry) for consistency.

**[LOW]** `is_path_like` requires both `/` and an extension, so `config/Dockerfile` (no extension) and Windows `\`-separated paths fall through to suggestions instead of resolving. Accept `\\` as a separator and treat the extension as a tightener, not a hard gate.

jahala (commented):
Reviewed by Opus 4.8:

Stepping back from the line-level notes to a direction question, since you've got the real-world context here.

This PR is really two features: (1) **fuzzy suggestions on a miss** ("did you mean `src/foo.rs`?"), and (2) **auto-opening the best guess** when the gate is confident. We love (1) — deterministic, helpful, saves a round-trip without ever pretending. We're less sure about (2): auto-resolving means tilth can return *a different file's content than was asked for*, and tilth's whole value is being the precise layer an agent doesn't second-guess. The silent-wrong-file failure (the unique-match note above) is the exact "confidently wrong" case tilth exists to avoid, and it adds non-determinism + a dependency to our most fundamental operation.

A suggest-only design keeps ~all the value (agent gets candidates, picks, one cheap hop) while preserving determinism — and it moots the unique-match issue entirely.

But you've been running this for real: what's the case for auto-open over suggest-only? Is the round-trip saving big in practice, or are there workflows where suggestions aren't enough? Genuinely want your read before we decide which way tilth leans here.

jahala:
Opus 4.8: @paulnsorensen — circling back on this one. The implementation is solid and every line-level note from the earlier review is addressed; the only open thread is the design question, and I want to settle direction with you before merging.

I lean **suggest-only**: when a path is slightly off, tilth returns the ranked candidate(s) and lets the caller open the one it wants — rather than auto-opening the top match and returning its contents directly.

The reasoning: tilth's core value is being a precise layer an agent doesn't second-guess. Auto-open introduces a new failure mode — tilth can hand back a *different* file's contents than was requested, and an agent may act on it without noticing the substitution. The round-trip that auto-open saves is cheap; that silent-wrong-file risk isn't. Suggest-only keeps all the path-intelligence benefit (one cheap hop to the right file), drops the unique-match/floor edge cases we went back and forth on, and is less code. If we go that way, the gate/scoring/security work you've done stays intact — it's essentially removing the `Resolved` auto-open branch and its callers.

What do you think about this?

jahala:
Maintainer action + decision. Two things landed on the branch: the overdue merge with main (one real conflict in tool_read, resolved keeping both features' logic), and the suggest-only conversion the review thread leaned toward — auto-open is removed, a confident match now renders as the top `did you mean` suggestion, verified identical across single read, batch read, and the search fallback. All gates green.

**Decision: HOLD — banked, not merged, on relevance evidence.** We went looking for measured demand before merging and found the opposite: across 2,892 tool calls in our benchmark corpus, wrong-path reads don't register as a failure category (≤0.3%, inside "other errors"). The dead-ends that actually cost turns are symbol misses (fixed via #170) and zero-match searches. Adding a permanent dependency (~550 lines + nucleo-matcher in the default build) for a sub-1% event doesn't clear our bar right now.

What would change this: evidence that path-misses matter outside short benchmark sessions — long real-world sessions with stale paths are the plausible case, and if telemetry or user reports show it, this branch is ready to land as-is. The implementation is good work; the demand evidence just isn't there yet.

jahala:
Closing with the final evidence, and with the work preserved. We measured real-world demand across 5+ months of actual session history (104k tool calls, organic sessions split from synthetic ones): path-misses that a fuzzy suggestion would have rescued run at 0.34% of reads in organic use — real, exactly the wrong-dir/typo shapes this PR targets, but low-frequency, and tilth's shipped did-you-mean already covers part of that. Below our bar for adding a dependency and ~550 lines to the default build.

The branch is the best version of this idea: rebased onto current main, suggest-only per the review consensus, all gates green — merge-ready as it sits. If future telemetry or user reports show path-misses mattering more than our data does (long sessions with stale paths remain the plausible case), this reopens and lands with no further work. Thank you for the port — the implementation quality was never the question.

PR #144, opened 2026-06-13
ci: dual-publish npm wrapper as @plotplot/tilth

jahala:
## What
Dual-publish the npm wrapper under the new `@plotplot` org scope alongside the canonical `tilth` package, so every release anchors the org with real code.

## Why
Owning the `plotplot` org already blocks external squatting; dual-publishing **reserves the canonical `@plotplot/tilth` name**, anchors the org with live code, and keeps a migration path open.

## How
- `publish-npm` job: after publishing `tilth`, rename the same `./npm` artifact to `@plotplot/tilth` and `npm publish --access public`.
- Both names use **OIDC trusted publishing** (no token). `@plotplot/tilth` has its own trusted publisher (`jahala/tilth` + `release.yml`) on npmjs.com.
- `@plotplot/tilth@0.9.0` was **bootstrapped** with a one-time manual publish — npm can't configure trusted publishing for a package that doesn't exist yet.

## Safety
- `version-check` unaffected — committed `package.json` stays `name: tilth`; the scope rename is ephemeral in CI.
- Unscoped `tilth` publishes **first**, so a scoped-step failure can't block the primary release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -139,12 +139,12 @@ jobs:
       - run: npm publish ./npm
       # Anchor the @plotplot org: dual-publish the SAME artifact as `@plotplot/tilth`.
       # Requires a trusted publisher for @plotplot/tilth configured on npmjs.com (one-time).
-      # Recovery: if this scoped publish fails AFTER the unscoped publish (above) has
-      # succeeded, re-running the job re-fails at `npm publish ./npm` (npm 403 — version
-      # already published) and never reaches this step. Recover by publishing the scope
-      # once manually — `cd npm && npm pkg set name=@plotplot/tilth && npm publish --access public`
-      # — or by cutting a new patch version.
+      # Best-effort: the unscoped `tilth` publish above is the canonical release, so a
+      # failure here (e.g. the @plotplot trusted publisher isn't configured yet) must not
+      # fail the release. continue-on-error keeps the job green; if it fails, publish the
+      # scope manually when ready — `cd npm && npm pkg set name=@plotplot/tilth && npm publish --access public`.
       - name: Dual-publish under the @plotplot scope
+        continue-on-error: true
         working-directory: ./npm
         run: |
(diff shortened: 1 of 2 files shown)


paulnsorensen (commented):
**Review (`/age`, ten dimensions)** — tight, correct 10-line release change. Unscoped-first ordering, `--access public` for the scoped name, and the per-package OIDC trusted-publisher model are all correct (verified against npm docs). One low-severity operability nit.

### Low

- **[ops · low]** `.github/workflows/release.yml:138-146` — partial-publish has no idempotent recovery. If the scoped publish (140-146) fails after the unscoped `npm publish ./npm` (138) succeeds, re-running the job re-fails on line 138 (`npm 403`, version already published), so the scoped step is never reached. Recovery is then the manual one-liner (`npm pkg set name=@plotplot/tilth && npm publish --access public`) or a new patch version. *Optional:* document the recovery in the step comment, or guard each publish with a "version already on registry" pre-check so re-runs are idempotent. (The gap pre-exists in the single-publish design; this PR extends it to the new step.)

Otherwise good — the canonical `tilth` release is protected by ordering, as the PR notes.

jahala:
Review by Opus 4.8:

Good call on the partial-publish recovery gap. Documented it in `49851a3`: a comment on the scoped step now spells out that a failure here *after* the unscoped publish succeeds makes a job re-run 403 at `npm publish ./npm`, with the manual recovery path (`cd npm && npm pkg set name=@plotplot/tilth && npm publish --access public`) or a new patch version. I left the idempotent "already on registry" pre-check as a follow-up rather than expand this PR, since — as you noted — the gap pre-exists in the single-publish design.

One operational prerequisite before the next tag: the `@plotplot/tilth` trusted publisher has to actually be live on npmjs.com or this step 403s on a real release. The CLAUDE.md note says it was bootstrapped — worth confirming that trusted-publisher config exists before relying on it. Not merging.

PR #151, opened 2026-06-19
feat(lang): add Bash tree-sitter support

jahala:
## Summary

Adds full Bash language support via `tree-sitter-bash` (0.25, ABI-compatible with tilth's 0.26 core) — a common scripting language that had no prior coverage.

**Detection:** `.sh` / `.bash` / `.bats` extensions, plus `.bashrc` / `.bash_profile` / `.bash_aliases` / `.profile` init files.

**Outline:**
- Functions in both syntaxes — `foo() { … }` and `function foo { … }` (reuses the existing `function_definition` path)
- Top-level variables — bare `VAR=…` plus `export` / `declare` / `readonly` declarations
- Function-local `local` and nested definitions are correctly excluded (top-level only)

**Search / grok:** callee extraction from command invocations (calls inside `$(…)` and pipelines resolve correctly); `source` / `.` import resolution; noise-stripping that keeps shebangs and strips only `DEBUG`-labelled `echo`.

Also refreshes the stale "Languages supported" list in CLAUDE.md (Scala/Kotlin/Elixir were already shipped but unlisted).

## Testing

- New in-source tests covering both function syntaxes, declarations, hyphenated names, callees, and import-source parsing.
- `cargo clippy -- -D warnings`, `cargo fmt --check`, `cargo test` (477 lib + 4 bin) all green on stable 1.96.
- Verified end-to-end on real scripts (outline, callers, grok) and stress-tested for panic-safety on adversarial input (unicode identifiers, malformed/unclosed syntax, deeply-nested subshells, associative arrays).
- Added Bash to the outline fuzz target.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -64,8 +64,8 @@ AGENTS.md              User-facing copy of the MCP instructions. Generated from 
 
 ## Languages supported
 
-Rust, TypeScript, TSX, JavaScript, Python, Go, Java, C, C++, Ruby, PHP, C#, Swift.
-Kotlin, Dockerfile, Make detected but have no tree-sitter grammar (outline returns None).
+Rust, TypeScript, TSX, JavaScript, Python, Go, Java, Scala, C, C++, Ruby, PHP, C#, Swift, Kotlin, Elixir, Bash.
+Dockerfile, Make detected but have no tree-sitter grammar (outline returns None).
 
 ## Build, test, install
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -864,6 +864,7 @@ dependencies = [
  "thiserror",
  "toml",
  "tree-sitter",
+ "tree-sitter-bash",
  "tree-sitter-c",
  "tree-sitter-c-sharp",
  "tree-sitter-cpp",
@@ -932,6 +933,16 @@ dependencies = [
  "regex-syntax",
  "serde_json",
  "streaming-iterator",
+ "tree-sitter-language",
+]
+
+[[package]]
+name = "tree-sitter-bash"
+version = "0.25.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9e5ec769279cc91b561d3df0d8a5deb26b0ad40d183127f409494d6d8fc53062"
+dependencies = [
+ "cc",
  "tree-sitter-language",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ tree-sitter-c-sharp = "0.23"
 tree-sitter-swift = "0.7"
 tree-sitter-kotlin-ng = "1.1"
 tree-sitter-elixir = "0.3"
+tree-sitter-bash = "0.25.1"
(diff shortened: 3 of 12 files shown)


paulnsorensen (commented):
**Review (`/age`, ten dimensions)** — clean, well-tested, additive. CI green (480 tests, clippy clean), no behavioral risk to the existing languages. Two optional low-severity nits, nothing blocking.

### Low

- **[correctness · low]** `src/lang/outline.rs:285-289` — a top-level subscript assignment like `ARR[0]=x` surfaces in the outline (and, via `DEFINITION_KINDS`, the symbol index) as `ARR[0]` rather than `ARR`. The bash grammar's `variable_assignment.name` field can be a `subscript` node, and `find_child_text` returns its full text verbatim. Cosmetic. *Optional:* when the name child is a `subscript`, extract its inner `variable_name`.
- **[correctness · low]** `src/read/imports.rs:66` — `is_import_line` / `extract_import_source` match only space-separated `source` / `.`. A tab-separated `source` builtin (e.g. `source⇥./lib.sh`), which is valid bash, is missed for deps analysis. Best-effort hint, so a miss only costs a related-file hint. *Optional:* split on the first run of whitespace.

### Verified non-issue (noting for confidence)
Adding `variable_assignment` to the shared, language-ungated `DEFINITION_KINDS` / `definition_weight = 60` is **bash-only in practice**. I checked `node-types.json` for all 16 other supported grammars — none emit `variable_assignment` (they use `assignment` / `assignment_expression` / `augmented_assignment` / etc.). So there is zero behavioral change for non-bash files; the only residual is latent — a future grammar reusing that node kind would inherit the weight silently.

### Nice touches

- `resolve_bash` uses a single `metadata()` instead of `exists()` + `is_file()` — avoids the TOCTOU.
- DEBUG-only `echo` stripping (keeps functional output), shebang preservation.
- The callee test derives `main`'s range from the outline, so it won't silently drift if the fixture is edited.

jahala:
Review by Opus 4.8:

Both nits fixed test-first in `1ffba48`:

- **subscript names** — `ARR[0]=x` now surfaces as `ARR`. Added an `assignment_name` helper that unwraps a `subscript` name node to its base `variable_name`, wired into both the `variable_assignment` and `declaration_command` bash arms. Confirmed `ARR[0]=hello` does parse as a top-level `variable_assignment` whose `name` field is a `subscript`, so the fix matches the real parse tree rather than guessing.
- **tab-separated `source`** — both `is_import_line` and `extract_import_source` now accept any whitespace (space or tab) after `source`/`.`, while still rejecting `sourcefile=1`, `./script.sh`, and `.bashrc`.

Failing-first tests added for all three (subscript outline, tab `source` extraction, tab `is_import_line` + the false-positive guards); local `cargo test` is green (480 passing). Leaving CI to confirm clippy on the stable toolchain.

Also took your point on the language-ungated `variable_assignment` weight — extended the `DEFINITION_KINDS` comment to flag the latent cross-grammar inheritance so a future grammar reusing that node kind is on the record. Not merging — over to you.

PR #153, opened 2026-06-22
fix(budget): guard rfind('\n') fallback to prevent empty truncation body

paulnsorensen:
`budget::truncate` could collapse a body with no interior newline to empty output. When the body is a single long line, the `rfind('\n')` fallback finds only the leading separator at offset 0; cutting there yields an empty body — a header plus a "... truncated" marker with zero content.

This guards that fallback arm with `.filter(|&p| p > 0)` so it falls through to `safe_max` and preserves content. The other two cut arms (`rfind("\n\n##")`, `rfind("\n\n")`) already carried this guard; this closes the one remaining hole. Adds a regression test asserting the truncated single-long-line body is non-empty.

Also adds doc comments to reserved `ViewMode`/`OutlineKind` variants and drops a now-stale `#[allow(dead_code)]` on `OutlineKind::Property` (constructed at `lang/outline.rs:222`).

Ported from paulnsorensen/tilth (fork PR #67).

### DIFF
--- a/src/budget.rs
+++ b/src/budget.rs
@@ -92,7 +92,7 @@ mod tests {
     }
 
     #[test]
-    fn apply_with_info_single_long_line_body_does_not_collapse() {
+    fn single_long_line_body_does_not_collapse() {
         // Regression: a body that is one very long line with no interior \n has no
         // `\n\n` to land on, so rfind('\n') finds only the leading separator at
         // offset 0. Without the `filter(|&p| p > 0)` guard on that arm, cut_point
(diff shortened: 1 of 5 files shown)


jahala:
Review by Opus 4.8:

**Approve — merge-ready.** The core fix is correct, minimal, and properly tested.

Traced it through `budget::apply`: `body = &output[header_end..]` always starts with the header/body separator `\n` at offset 0, so for a single-long-line body the only `\n` in `truncated` is that leading one — `rfind('\n')` returns `Some(0)`, `cut_point` becomes 0, and `clean_body` collapses to empty. Adding `.filter(|&p| p > 0)` falls through to `safe_max`, and since `safe_max = body.floor_char_boundary(max_bytes)` the `&body[..safe_max]` slice stays UTF-8-safe — the new path can't panic on multibyte content. Good consistency, too: the other two cut arms already carried `p > 0`; this closes the last hole. The regression test is a genuine test-first guard (asserts an `'x'` survives, which fails on the old code).

Confirmed the `#[allow(dead_code)]` drop on `OutlineKind::Property` is right — it's constructed at `lang/outline.rs:222` and rendered at `read/outline/code.rs:174`, so the allow was stale and slightly misleading.

Two optional notes, neither blocking:
- **Scope** — the doc/comment tidies in `lib.rs`/`map.rs`/`overview.rs`/`types.rs` are unrelated to the budget fix (and you disclosed them). All harmless and a net improvement, but bundling them under a `fix(budget)` title reads slightly less cleanly in history; splitting incidental cleanups is a nice-to-have, not a request.
- **For completeness** — a budget at or below `header_reserve` (50 tokens) still collapses the body, because `max_bytes` → 0 forces `safe_max` to 0. Pathological (the default is 24k) and out of scope here — noting only so it's on record.

Nice catch on a subtle one.

PR #154, opened 2026-06-22
fix(diff): surface git errors, core.quotePath=false, fail-closed overlays

paulnsorensen:
Hardens structural-diff error handling and edge cases (four changes):

1. `run_git_diff` and `git_show` pass `-c core.quotePath=false`, so non-ASCII paths are no longer octal-escaped.
2. Conflict-formatting output is budgeted independently and `budget::apply` is called once on the combined string, preventing roughly 2× cap overage when a large conflict block is present.
3. All four overlay builders (`compute_modified` / `compute_renamed` / `compute_added` / `compute_deleted`) now fail closed: a `git_show` or filesystem error yields an "unavailable" overlay with no symbol changes instead of a silently-wrong all-Added / empty result.
4. `parse_hunk_header` returns `None` on an absent or non-numeric start spec, so malformed hunks are skipped rather than mis-bucketed at line 0.

Adds regression tests for the bare/range git-ref new-content paths, the malformed-hunk skip, and git_show-error handling.

Ported from paulnsorensen/tilth (fork PR #69).

### DIFF
--- a/src/diff/matching.rs
+++ b/src/diff/matching.rs
@@ -23,6 +23,51 @@ pub(crate) fn build_diff_symbols(
     out
 }
 
+/// Pair entries within one identity bucket (same name+kind — i.e. overloads).
+/// Exact content-hash twins are paired first regardless of position, so an
+/// unchanged overload matches its real counterpart even after reordering or
+/// insertion; the remainder pair in positional order. Old/new entries left
+/// over (unequal counts) are simply absent from the result and fall through
+/// to the structural/fuzzy phases.
+fn pair_overloads(
+    old: &[DiffSymbol],
+    new: &[DiffSymbol],
+    old_indices: &[usize],
+    new_indices: &[usize],
+) -> Vec<(usize, usize)> {
+    let mut pairs = Vec::with_capacity(old_indices.len().min(new_indices.len()));
+    let mut new_used = vec![false; new_indices.len()];
+    let mut old_unpaired: Vec<usize> = Vec::new();
+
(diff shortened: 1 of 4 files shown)


jahala:
Review by Opus 4.8:

**Request a small change, then good to merge.** Three real bugs fixed cleanly with locked tests: the bare-`GitRef` new side now reads the working tree instead of `HEAD:<path>` (`overlay.rs` `resolve_git_ref_new_side`) — a genuine content mis-attribution; `parse_range_spec` returns `Option` instead of `unwrap_or(0)`; `pair_overloads` two-pass content-hash-then-positional. `core.quotePath=false` is applied to both `git diff` and `git show` but not the fs-read conflict path — exactly right.

**Change before merge** — `src/diff/mod.rs:476`: `diff_log` still does `run_git_diff(&commit_source).unwrap_or_default()`, swallowing git errors in the per-commit log path that the rest of this PR sets out to surface. Either propagate per-commit (skip with a note) or add a comment marking it intentional best-effort, so it doesn't read as an oversight against the PR's own goal.

**Nit** — `src/diff/overlay.rs:346`: the `get_old_content` doc comment is truncated mid-sentence ("…the caller should").

paulnsorensen:
Addressed in `dac5be7`:

- **`diff_log` per-commit path** now propagates git errors via `?` instead of `.unwrap_or_default()`, matching the fail-closed `?`/`Err` handling already used for the `git log` step in the same function. A real git failure now aborts the log rather than silently emitting an empty diff.
- Completed the truncated doc comment on `get_old_content` in `overlay.rs`.

PR #155, opened 2026-06-22
fix(cache,edit,install): bounded cache eviction + atomic file writes

paulnsorensen:
Two durability/correctness fixes plus a small TOML-scan fix:

1. **Bounded cache eviction.** `OutlineCache` is rekeyed from `(PathBuf, SystemTime)` to `PathBuf`, with the mtime moved into the value. Inserting a new mtime now replaces the prior entry in place, bounding the cache to one entry per live path — the old keying never evicted stale-mtime entries, so the cache grew unbounded as files changed. The lookup reference is also scoped to its block so no `DashMap` reference is held across the insert (defensive hardening of the guard lifetime).
2. **Atomic file writes.** Writes in `edit` and `install` go through a shared `util::atomic_write_bytes`: write to a collision-free temp file in the same directory, copy the original file's permissions onto the temp, then atomically rename over the target. A crash mid-write can no longer leave a truncated file.
3. **TOML section scan.** `toml_section_end` uses a bracket-depth scan that is correct for multibyte content, comments, array-of-tables headers, and single-quoted strings, replacing a bare `"\n["` search.

Adds tests for stale-mtime eviction and the TOML value-with-bracket case.

Ported from paulnsorensen/tilth (fork PR #70).

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -2,13 +2,14 @@ use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::time::SystemTime;
 
-use dashmap::mapref::entry::Entry;
 use dashmap::DashMap;
 
 use crate::types::Lang;
 
-/// Cached outline entry.
+/// Cached outline entry keyed by path. `mtime` is stored in the value so a
+/// stale entry is detected and evicted in O(1) (one map lookup, no `retain`).
 struct CacheEntry {
+    mtime: SystemTime,
     outline: Arc<str>,
 }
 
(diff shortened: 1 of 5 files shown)


jahala:
Review by Opus 4.8:

**Two small changes requested.** The cache key redesign from `(PathBuf, SystemTime)` to `PathBuf`-with-mtime-in-value is the correct fix for unbounded growth (O(files), not O(edits)), and the eviction test uses a panicking closure to prove the fast path is taken. Atomic write = temp-in-same-dir + rename with best-effort cleanup; `toml_section_end` is a sound replacement for the fragile `find("\n[")` heuristic.

**Change before merge**
- `src/install.rs:205,248` — `depth: i32` is decremented on every `]` with no floor; an unmatched `]` in a hand-edited config drives it negative and can truncate the section being rewritten. `depth = depth.saturating_sub(1)`.
- `src/util.rs:15` — `path.parent().unwrap_or_else(|| Path::new("."))`: `Path::new("config.toml").parent()` is `Some("")`, not `None`, so a bare filename lands the temp file in the cwd; if the cwd is on a different mount than the target, `rename` fails `EXDEV`. Dormant today (callers pass absolute paths) but a real trap — `.filter(|p| !p.as_os_str().is_empty())`.

**Nits** — note the `get_or_compute` double-compute window (two stale readers both recompute — safe, just redundant vs the old `entry()` API); a one-word comment on the `Relaxed` counter (uniqueness only).

paulnsorensen:
Addressed in `815ad80`:

- **`toml_section_end`**: `depth` is now `u32` with `saturating_sub(1)`, flooring at 0. Previously an unmatched `]` drove the `i32` depth negative, so the `depth == 0` header-boundary check never re-fired and the rewrite truncated the rest of the file. Added a regression test through `upsert_toml_section` asserting a following `[other]` table survives.
- **`atomic_write_bytes`**: filter the empty-string parent that `Path::new("foo.txt").parent()` returns (`Some("")`) so the temp dir is an explicit `.`. This is a defensive normalization — `Path::new("").join(name)` already resolves to the cwd, so behavior is unchanged, but the explicit `.` documents intent and guards future refactors.

PR #156, opened 2026-06-22
fix(search): Go stdlib classification, fixture penalty, footer dedup

paulnsorensen:
Four search-ranking / classification fixes:

1. Go `is_stdlib` matches the first `/`-segment of an import path against `GO_STDLIB_ROOTS`, so `net/http` and `encoding/json` classify as stdlib; adds `cmp` / `maps` (Go 1.21+) and drops the spurious `http` / `utf8` roots.
2. `fixture_penalty` is anchored to a path *component* rather than a substring, so a file like `examples_parser.rs` is no longer penalized as a fixture.
3. `multi_word_boost` tokenizes the query on the same boundary rule as the haystack, so multi-word queries containing punctuation match symmetrically.
4. The callers hop-2 footer reports the uncapped unique total rather than the capped display count.

Note vs the fork original: the fork's `.tilthignore`-only walker change to `find_basename_fallback` is intentionally omitted — this upstream has no `.tilthignore` feature, so the existing ignore-source walker is kept unchanged.

Ported from paulnsorensen/tilth (fork PR #71).

### DIFF
--- a/src/search/callers.rs
+++ b/src/search/callers.rs
@@ -418,15 +418,11 @@ pub fn search_callers_expanded(
                     );
                 }
 
-                let _ = writeln!(output, "\n{} functions affected across 2 hops.", {
-                    // Dedup hop-1 by calling_function for a distinct-function count.
-                    let hop1_fns: std::collections::HashSet<&str> = sorted_callers
-                        .iter()
-                        .filter(|c| c.calling_function != "<top-level>")
-                        .map(|c| c.calling_function.as_str())
-                        .collect();
-                    hop1_fns.len() + unique_total
-                });
+                let _ = writeln!(
+                    output,
+                    "\n{} functions affected across 2 hops.",
+                    all_caller_names.len() + unique_total
+                );
             }
         }
     }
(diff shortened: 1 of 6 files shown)


jahala:
Review by Opus 4.8:

**One change requested, then good to merge.** The Go stdlib prefix-match fix with eight regression-named tests is the cleanest test block in this batch, and the fixture-penalty move to component-anchored matching is the right idiom.

**Change before merge** — `src/search/callers.rs:421-428`: the "N functions affected across 2 hops" footer rebuilds `hop1_fns` from `sorted_callers` *after* `truncate(max_matches)` (line 315), so with more than `max_matches` hop-1 callers the count silently caps (10 by default). You already collect `all_caller_names` pre-truncation (line 309, comment "for accurate fan-out threshold") — reuse `all_caller_names.len()` and drop the inner `hop1_fns` rebuild.

**Nits** — `GO_STDLIB_ROOTS` is missing `iter`/`unique` (Go 1.23), so `import "iter"` classifies as external; inverting the rule to "first path segment contains `.` ⇒ external" would end the allowlist drift entirely. And `rank.rs` `non_code_penalty` still uses substring `contains("example")` — the same false-positive pattern you fixed in `fixture_penalty`; anchor it to a path component too.

paulnsorensen:
Addressed in `5547986`:

- **Hop-2 footer count**: the "N functions affected across 2 hops" line now uses `all_caller_names.len()` (the distinct hop-1 caller set collected *before* `truncate(max_matches)`) instead of rebuilding from the truncated `sorted_callers`. With more than `max_matches` hop-1 callers the count was silently capped; it now reflects the true total. Dedup semantics are unchanged (distinct `calling_function`, excluding `<top-level>`). Regression test `footer_count_uses_pre_truncation_caller_set`: 8 callers × 2 call sites + 1 hop-2 caller → footer reports exactly 9.
- **`non_code_penalty`**: anchors the example/sample/template check to a path component (mirroring the `fixture_penalty` fix already in this PR), with tests for substring-not-penalized and component-penalized.
- **`GO_STDLIB_ROOTS`**: added `iter` and `unique` (Go 1.23 stdlib packages).

PR #75, opened 2026-05-07
feat: pull Tier-1 standalone fixes from @choongng/dev

jahala:
Massive thanks to @choongng for the substantial work on the [dev branch](https://github.com/choongng/tilth/commits/dev/) — this is the first of several PRs pulling that work into main with full author attribution preserved.

This PR cherry-picks 7 standalone fixes (no inter-dependencies, no behavior changes beyond the bug-fixes themselves).

## Commits (Choong Ng as Author on each)

| SHA on dev | Title |
|---|---|
| `4fa6832` | Detect terminal height via TIOCGWINSZ; stop paging based on stale 24-row fallback |
| `d8c9617` | Ensure CLI output ends with a newline |
| `87b0b94` | Drop OutlineKind::Export doubling; recurse into wrapped declaration |
| `8b585f4` | index: drop inert SymbolIndex plumbing (-547 LOC of dead code) |
| `5d81660` | mcp: drop dead tilth_map dispatch arm |
| `da5c954` | imports: normalize resolved paths so different spellings dedup |
| `6c3f3b3` | fix: detect JS/TS export const and bare const as definitions |

## Our additions

- **`3f0b377`** — pinned regression test (`ts_export_outline_no_doubled_keyword`) for the export-doubling bug. Choong's fix had no regression test; this commit pins the post-fix behavior so the bug can't silently return.
- **`f5bc8cc`** — clippy + fmt drift cleanup. `da5c954` tripped `clippy::unnested-or-patterns` and `6c3f3b3` had a long `assert!` line that rustfmt re-wraps; not caught on choongng/dev because their toolchain is older.

## Notable changes

- New dependency: `terminal_size = "0.4"` (used by 4fa6832 for proper TTY height detection on systems without `\$LINES`)
- 547-line dead-code removal in 8b585f4 — `SymbolIndex` was allocated and threaded through but every consumer had `let _ = index;` discarding it
- 6c3f3b3 fixes a real bug: `extract_definition_name` was probing wrong fields on `lexical_declaration` (TS) and `variable_declaration` (C#), missing `export const Foo = ...`
- 87b0b94 fixes a visible outline rendering bug: `export async function foo() {}` was outlining as `"export export async function foo("`

## Test plan

- [x] `cargo test --lib` — 267 passed (was 266; new test added)
- [x] `cargo clippy -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
[...]

PR #157, opened 2026-06-22
fix(grok): resolve qualified targets (Type::method) and soften MCP instruction wording

paulnsorensen:
Two changes to `tilth_grok`:

1. **Qualified-target resolution.** `Type::method` / `Type.method` now resolves to the underlying method definition (asserting the exact start_line) instead of returning NotFound. A literal lookup is tried first; only on a miss does a bounded trailing-segment retry resolve the bare method name, so bare-symbol resolution is unchanged.

   Limitation: when several types define a method of the same name, the retry resolves to the top-ranked same-named definition without honoring the qualifier (e.g. `TypeA::method` may resolve to `TypeB::method`). This is strictly additive — these targets previously returned NotFound — and `other_def_count` surfaces the ambiguity to the caller. Full owner-disambiguation is intentionally out of scope for this change.

2. **Instruction wording.** Softens absolutist phrasing in the MCP server instructions (`prompts/mcp-base.md`, regenerated into `AGENTS.md`); the embedded `SERVER_INSTRUCTIONS` byte-lock is updated to the new prompt bytes.

Ported from paulnsorensen/tilth (fork PR #75). The underlying minimal resolver retry originated in fork commit `2f7c448` (outside the original PR); without it the regression test cannot pass on this base. That commit's broader owner-disambiguation machinery is not included here.


jahala:
Review by Opus 4.8:

**Approve the grok fix; requesting changes on the prompt hunk.** `Type::method` resolution via a bounded `rfind` retry that leaves the bare-symbol path untouched is clean — nested qualifiers and `Foo<T>::bar` generics both work, and the byte-lock (`3466→3471`) + AGENTS.md regen are in lockstep.

**Change before merge** — `prompts/mcp-base.md` (top cluster) downgrades `DO NOT use Grep, Read, or Glob.` and the imperative `use X instead of Y` lines to soft `prefer X over Y`. CLAUDE.md explicitly records that hard `DO NOT` framing outperforms soft wording for weaker models, and that instruction changes must be validated with haiku benchmarks on the hard tasks (`rg_search_dispatch`, `rg_trait_implementors`, `gin_servehttp_flow`) — this is the most model-sensitive region of the prompt, and there's no benchmark attached. Please land the grok fix now and either split the prompt change into its own benchmarked PR or attach haiku numbers showing no regression.

**Should fix** — add unit tests for `split_qualified` (nested `A::B::method`, generics, bare round-trip) and the not-found path (`Type::nonexistent` → `NotFound`).

jahala:
**Reviewed by Opus 4.8:**

# Review — PR #157: resolve qualified grok targets + soften MCP instruction wording

**Verdict: the grok change is good and mergeable on its own; the instruction-wording change needs a benchmark gate; and as it stands this PR carries two unrelated layers underneath it.** I'd split the grok fix out.

## What I verified
- `cargo fmt`, `cargo clippy -- -D warnings` (CI-exact), and the full suite pass.
- This is the **top of a 3-PR stack**: it contains *all* of #158 and #159. So merging #157 ships the read-side hard-refuse (#159) and the write-side `root` anchoring (#158) too — including #158's blocking bug (see that review: `root` alone doesn't enable a cross-worktree write). Those issues gate this PR transitively.

## Part 1 — qualified-target resolution (`Type::method`): good

The implementation is clean and correctly scoped:
- Literal lookup fires first, so bare-symbol resolution is genuinely unchanged (`bare != name` is false for a bare target, and the literal match returns before `split_qualified` is consulted).
- The trailing-segment retry is bounded — it only fires when the literal lookup found no definition.
- The documented limitation (when several types define a same-named method, the retry resolves to the top-ranked one without honoring the qualifier) is surfaced to the caller via `other_def_count`. Acceptable and clearly noted.

Audited for robustness — no issues:
- `split_qualified`'s `rfind([':', '.'])` indexes a 1-byte ASCII separator, so `&name[idx+1..]` is always on a char boundary even for multibyte targets (`café.método`, `日本.語`, `😀.x` — no panic).
- Degenerate inputs `method:`, `..`, `a.`, `:` yield `""`, which the `!bare.is_empty()` guard skips.
- **[NOTE]** a filename-shaped target like `foo.rs` splits to `rs` and, only if no definition is literally named `foo.rs`, retries a symbol search for `rs` — which could spuriously resolve to an unrelated symbol named `rs`. Bounded (returns `NotFound` if `rs` also has no definition), no panic, low probability. Worth a one-line guard if you want to be strict (e.g. skip the retry when the bare segment looks like a file extension), but not blocking.

If you split this part into a standalone PR off `main`, it could merge quickly — it's independent of the `root` work.

## Part 2 — softening the MCP instruction wording: benchmark first

This rewrites the strong steering:

```
- DO NOT use Grep, Read, or Glob. Always use the better tools …
+ For file reads, text search, and directory listings, prefer tilth_search, tilth_read, and tilth_files over Grep, Read, or Glob.
```
…and similar "use X instead of Y" → "prefer X over Y" changes, plus the byte-lock update.

This runs directly against the project's own documented guidance: **"DO NOT" framing beats soft framing for weaker models**, and instruction changes are haiku-sensitive and must be benchmarked on the hard tasks (`rg_search_dispatch`, `rg_trait_implementors`, `gin_servehttp_flow`) before merge. Softening "DO NOT use Grep" to "prefer … over Grep" is exactly the kind of change that can quietly reduce haiku's tilth adoption — which shows up as higher cost-per-correct, not as a test failure. There's no evidence in the PR that these benchmarks were run.

**Ask:** run the haiku benchmarks baseline-vs-this-branch on those three tasks and post the tilth-tool-call counts / cost-per-correct. If adoption holds, merge the wording; if it drops, keep the imperative framing. The byte-lock test was updated correctly and `AGENTS.md` was regenerated in lockstep — that part is fine.

## Recommendation
1. **Split the grok resolver** into its own PR branched off `main` — it's good, independent, and shouldn't be held hostage by the `root` stack.
2. **Gate the wording change** on the haiku benchmark numbers.
3. Resolve #158's blocking bug and #159's design decision before any of this stack lands, since #157 inherits both.

## Concurrency / correctness audit
No blocker- or major-level concurrency, TOCTOU, or panic issues found across the full stack (details in the #158 and #159 reviews). Dispatch is sequential, the write boundary is robust against `..`/symlink traversal, writes are atomic, and all new `unwrap`/indexing is confined to test code.

jahala:
Maintainer action: the grok resolver — the half this review approved standalone — is now split onto main as #170, with your authorship preserved on the resolver commit and the requested `split_qualified` unit tests added. It also picked up independent evidence since the review: our turn-anatomy study measured qualified-name grok misses as the single biggest dead-end source in the benchmark corpus (100 calls, 3.5% of all turns), so thank you — this was the right fix to port.

That leaves this PR carrying only the instruction-wording change (DO NOT → prefer), which sits against the repo's documented, measured finding that hard framing outperforms soft framing for the Haiku target — and no benchmark numbers have been posted. Unless evidence lands, the plan is to close this PR once #170 merges: the valuable half will have shipped, and the wording change shouldn't merge on taste against data. Happy to run the standard Haiku battery on it if you want to make the case.

jahala:
Closing now that the durable half has shipped: the qualified-target resolver merged via #170 with your authorship, and independent measurement confirmed it fixes the single largest dead-end source in our corpus. Good fix — thank you.

What remains here is only the instruction-wording change, which runs against the repo's measured finding on hard-vs-soft framing for the weakest target model, with no counter-evidence posted. The offer stands as an open door rather than a dead end: if you want to pursue it, we'll run the standard Haiku battery on the wording and let the numbers decide.

PR #158, opened 2026-06-22
fix(write): anchor tilth_write paths against an explicit root, not server cwd

paulnsorensen:
Adds a per-call `root` argument to `tilth_write` so relative target paths anchor under an explicit absolute directory instead of the server's process cwd. Without this, an MCP server launched from a different directory than the editing worktree silently wrote to the wrong place.

Behavior:

- Absolute target paths are used as-is.
- A relative target anchors under `root`. A `root` value that is not absolute is rejected.
- When `root` is omitted and a relative write resolves into a different git worktree than the server cwd, the write proceeds but a cross-worktree warning is appended — it does not hard-refuse, since the in-tree relative path is the dominant case.
- Success outputs (overwrite / append / hash modes) echo the resolved absolute path in the `##` header so a misplaced write is visible.

**BREAKING:** a non-absolute `root` value is rejected; relative target paths resolve against `root` when supplied.

Notes for review:

- Upstream's `apply_batch` returns only a combined string (no written-path list), so the hash-mode cross-worktree warning is computed over the call's tracked relative paths, filtered by on-disk existence. In a mixed-result hash batch it can advise on a cross-worktree path whose individual edit failed; the warning text stays accurate and nothing is mis-written.
- This is the write-side half of the `root` work. The read-side change (separate PR) makes the *stricter* choice of hard-refusing a relative read path/scope without `root`; writes deliberately warn-and-write instead. The asymmetry is intentional — a wrong-cwd read returns harmless wrong content, whereas refusing every relative write would break the common in-tree path.

Ported from paulnsorensen/tilth (fork PR #76).


jahala:
Review by Opus 4.8:

**Approve the approach — please hold for reconciliation with #159 rather than merge standalone.** Anchoring write paths to an explicit `root` instead of the server cwd is the right fix, and the write *confinement* is genuinely sound: `path_within_scope` canonicalizes (resolving symlinks) and `tool_write` fail-closed *refuses* out-of-scope writes for absolute and relative paths alike. `EDIT_MODE_EXTRA` byte-lock (`2108→2457`) + AGENTS.md regen are correct.

**Change before merge** — `src/mcp/tools/write.rs`: the file path is anchored to `root`, but `scope` (the containment boundary) is still resolved against the server cwd via the old `resolve_scope(args)`. The two anchors should match, or a legitimate cross-worktree write (`root` ≠ server cwd) gets spuriously refused. #159 introduces `resolve_scope(args, root)` — adopt it here when this rebases on #159 (see merge note), so both PRs share one anchor.

**Nits** — the `cross_worktree_warning` block is duplicated verbatim in the overwrite and append arms (extract a helper); `hash_mode_cross_worktree_warning_fires` tests the helper directly rather than end-to-end through `tool_write` (the overwrite tests already show the two-tempdir-worktree pattern).

**Merge note** — #157/#158/#159 all touch `AGENTS.md` + `src/mcp/mod.rs` byte-lock + `tools/`. Suggested sequence: #159 first (it adds `anchor_path` + the `resolve_scope(args, root)` signature), then #158 rebased on top adopting that signature — which closes the scope/cwd gap above in one move.

jahala:
**Reviewed by Opus 4.8:**

# Review — PR #158: anchor `tilth_write` paths against an explicit root

**Verdict: do not merge yet — the headline feature does not work in its target scenario.** The `root` argument redirects where a path *resolves*, but the write-containment guard still pins to the server's process cwd, so a `root`-anchored write into a different worktree is **refused** before it happens. The very case this PR exists to fix (server launched in a different directory than the editing worktree) still fails — just loudly instead of silently.

## What I verified
- `cargo fmt`, `cargo clippy -- -D warnings` (CI-exact), and the full suite pass.
- This PR **contains all of #159** (read-side hard-refuse) plus the write-side change. It is not "write-side only" as the description implies — the read-side require-root refactor rides along. Merge after #159.

## Blocking issue — `root` alone does not enable a cross-worktree write

`root` flows into `resolve_write_path` (path resolution) but **not** into `scope_root` (the containment boundary), which still defaults to `current_dir()`:

```rust
let scope_root: PathBuf = match args.get("scope").and_then(|v| v.as_str()) {
    Some(s) => PathBuf::from(s),
    None => std::env::current_dir().unwrap_or_else(|_| ".".into()),
};
```

So when an agent does the intended thing — `tilth_write(root=/worktreeB, files=[{path:"src/x", mode:"overwrite", …}])` from a server whose cwd is `/checkoutA` — the path correctly resolves to `/worktreeB/src/x`, but `path_within_scope(/worktreeB/src/x, /checkoutA)` is false and the write is rejected. Reproduced directly:

```
## /worktreeB/out.txt
error: refusing write outside scope (/server/cwd)
→ file NOT created
```

The guard runs for **all** modes (hash/overwrite/append), so this isn't limited to direct writes. The only way to make it work today is to *also* pass `scope=/worktreeB` — but `scope` is **not documented in the write tool's schema** (only `files`, `diff`, `root` are), so a schema-following agent cannot discover it, and the error message doesn't mention it either.

**Root cause:** resolution and containment use two different anchors. They need to agree. The cleanest fix is to default `scope_root` to `root` when `root` is supplied (falling back to cwd only when neither is given), with care for the absolute-path case (an absolute target outside `root` should still be allowed per the schema's "absolute paths used as-is" contract — so the containment for absolute targets likely needs to derive from the target itself, not `root`). This deserves a small design pass rather than a one-liner.

**Please add a test for the root-only case** (`root` set, `scope` omitted, target relative, cwd ≠ root). Every current write-root test passes `scope` alongside `root`, which is exactly why this gap slipped through green.

## Security / TOCTOU audit (reassuring)

I specifically checked the write path for traversal and races; the boundary holds:
- **`..` traversal is refused.** `scope/sub/../../outside/x`, `../x`, and `ghost/../../x` (with `..` in a non-existent prefix) are all rejected — `canonicalize` normalizes any resolvable `..`, and `Path::file_name()` returning `None` on a trailing `..` bails the containment check to `false`.
- **Symlink escape is refused.** A symlink inside scope pointing outside is caught: `canonicalize` resolves it before the `starts_with` check, and the write primitives additionally open with `O_NOFOLLOW` (overwrite-rewrite, append) and `O_EXCL`/`create_new` (create-only).
- **Writes are atomic.** Hash mode uses temp-file + `rename`; create-only is `create_new(true)`. No exists()-then-write TOCTOU.
- **Concurrency:** tool dispatch is sequential (one in-flight call; overlap only via timed-out stragglers, which merely *read* cwd → benign). No data race introduced.

The remaining TOCTOU window between `path_within_scope` and the write is real but mitigated by the syscall-level `O_NOFOLLOW`/`O_EXCL` guards above; only an intermediate-directory symlink swap by a process that already has write access inside the canonical scope is theoretically uncovered. Low severity, and not introduced by this PR.

## Minor / notes
- **[MINOR] spurious cross-worktree advisory on failed hash edits (self-admitted, confirmed).** In hash mode, paths are pushed to `hash_relative_paths` *before* the edit runs, and `cross_worktree_warning` only self-filters paths that don't exist on disk. A hash edit that *fails* against a pre-existing file (e.g. `HashMismatch`, which does not delete the file) still canonicalizes, so a misleading "cross-worktree write" advisory is appended for a path whose edit failed. Advisory text only — no mis-write — but confusing in a mixed batch.
- **[NOTE] `find_git_root` has a latent bug for relative input** (a relative non-existent path can walk to `""` and test `.git` against the process cwd). Both production call sites pass only absolute paths, so it's currently unreachable — but the function isn't internally guarded.
- **[NOTE] `apply_one` success header shows the canonicalized (symlink-resolved) target.** Intentional per the "show where the write landed" goal, but an agent editing through a symlink will see the real target, not the path it passed.

## Things that are solid
The atomic-write machinery, the `O_NOFOLLOW` guards, the canonical-containment check, and the echo-resolved-absolute-path-in-the-header idea are all good. The feature concept is right; it just needs the containment anchor aligned with the resolution anchor before it actually delivers.

PR #79, opened 2026-05-08
docs: ARCHITECTURE.md from @choongng/dev (trimmed for durability)

jahala:
Adds a 700-line architectural review at `ARCHITECTURE.md`. Originally authored by @choongng on the dev branch — pulls 5 of their commits with author preserved (`b3fea19`, `850d7ff`, `cc0f0e4`, `5c9806d`, `5fb8fd1`), then layers a surgical edit on top.

Stacked on #77 (PR-D) so the structural facts the doc describes (e.g. `bloom_walk`/`callee_query`/`scope` modules, `read_ranges` multi-section reader) actually exist when this lands.

## What's in the doc

The durable architectural reasoning — type alphabet, dispatch flow, search/read/diff subsystem internals, caches, MCP threading model, extension-point cookbook, open architectural threads.

The most valuable section is "Open architectural threads" — captures real bugs/limitations (canonical-path symlink dedup, MCP per-request thread leak, OutlineCache eviction) that aren't documented anywhere else.

## What was trimmed (-75 lines, 931 → 856)

The agent audit found ~40% of the original was decaying ephemera. Trimmed:

- "Reading order" prescriptive list — opinion that decays.
- "Fork delta" table (62 lines) — a curated git log; rots every cherry-pick.
- ~25 specific LOC counts (`callers.rs ~900 lines`, `~22.1k Rust`, etc.). Three were already wrong relative to PR-D state when the doc was authored against `cb2d875`.
- All `(fork commit XXXX)` inline citations — couple prose to commit hashes that won't survive squash-merge.
- "Session 5N" internal references (fork-side framing).

## Synced to PR-D state

- search/ now lists 16 files (added `bloom_walk.rs`, `callee_query.rs`, `scope.rs` from `e3c4907` extraction).
- read/ describes `read_ranges` (multi-section) + `sections: [...]` array param from `4e25400`.
- Symbol search section explains `walk_to_enclosing_definition` lives in scope.rs post-extraction.
- Removed `callers.rs ~900 lines` claim — collapsed to ~540 by `2d5365b`.
- Added a paragraph on AGENTS.md / SERVER_INSTRUCTIONS sync discipline.
- Reframed intro to drop the `cb2d875` reference + "30 commits ahead" framing.

## Test plan

- [x] `wc -l ARCHITECTURE.md` = 856 (was 931 on choongng/dev)
[...]

PR #159, opened 2026-06-22
fix(mcp): require an absolute path or absolute root on read-side file-I/O tools

paulnsorensen:
Read-side MCP tools (`tilth_read` / `tilth_search` / `tilth_files` / `tilth_deps` / `tilth_grok`) resolved relative paths and scopes against the server's frozen launch cwd. Inside a git worktree that cwd is the wrong checkout, so a relative read silently hit the parent tree.

This makes the server stateless with respect to cwd:

- A relative path or scope now requires an absolute `root` and is hard-refused otherwise, with an actionable message.
- Absolute paths are used as-is; omitting `root` with absolute inputs is unchanged.
- `resolve_scope` becomes fallible and threads an `Option<&Path>` root; its missing-directory branch falls back to `root` or errors — never to the server cwd.

Factors shared `anchor_path` / `resolve_read_path` helpers, threads `root` through the five tools and their input schemas, inverts (does not delete) the old cwd-relative guards into per-tool refusal tests, leads the prompt guidance with the require-root directive, and regenerates AGENTS.md.

**BREAKING:** read-side tools refuse a relative path/scope when no absolute `root` is given.

Relationship to the write-side `root` change (separate PR):

- Write-side `root` anchoring is a sibling PR; this branch leaves `write.rs` behavior unchanged and stands alone on `main`.
- The two share an anchor helper (`anchor_path`) and the now-fallible `resolve_scope`; if both merge, expect a small dedupe touchpoint in `src/mcp/tools/mod.rs`.
- Reads hard-refuse a rootless relative path; writes deliberately *warn-and-write* instead. The asymmetry is intentional — a wrong-cwd read returns harmless wrong content, whereas refusing every relative write would break the common in-tree path.

Known scope boundary (matches the original): a `tilth_grok` target written as a file-prefixed location (e.g. `src/file.rs:7`) still resolves its file prefix against cwd; only the scope channel is root-anchored.

Ported from paulnsorensen/tilth (fork PR #79).


jahala:
Review by Opus 4.8:

**One doc fix requested, then good to merge.** The require-absolute-or-root discipline (shared `anchor_path`, consistent `resolve_scope(args, root)` across every read tool, breaking change locked by tests) is the right fix for the silent *wrong-worktree* read footgun. Worth stating plainly: this is a footgun fix, not a sandbox — read tools deliberately accept any absolute path, so there's no confinement boundary on reads (and none is intended).

**Change before merge** — `src/mcp/tools/definitions.rs:74` (and `:267`): `tilth_read`'s `path` still advertises "Absolute or relative file path to read." after this PR forbids relative-without-root. An agent following the schema will pass a relative path and hit the new error with no hint. Update it to say a relative path requires an absolute `root`.

**Should fix** — `src/mcp/tools/write.rs` carries the same scope-vs-cwd gap as #158 (reconcile on the #158 rebase).

**Nit (not a vuln)** — `anchor_path` does `root.join(raw)` without normalizing `..`, so `root=/x, path=../../y` resolves to `/y`. Since absolute reads are allowed by design, this reaches nothing the agent couldn't already name directly — but normalizing (or rejecting an escaping `..`) keeps the cross-worktree intent and error messages accurate; add a test pinning the behavior either way.

**Praise** — the content-assertion added to the byte-lock test (`contains("DO NOT pass a relative path/scope without an absolute root")`) is a stronger drift guard than a byte count, and the breaking-change tests name the *why*.

paulnsorensen:
Pushed the fix and brought the branch current with `main`.

**Change-before-merge (doc fix):** the `tilth_read` `path` schema at `definitions.rs:74` and `:267` now states that a relative path requires an absolute `root` (and that the server cannot see the shell cwd), so an agent following the schema no longer walks blind into the new relative-without-root error.

**`anchor_path` `..` nit:** added `anchor_path_dotdot_not_normalized`, pinning the current behavior (`root="/x"` + `"../../y"` → `/x/../../y`, unnormalized) and leaving behavior unchanged — so any future traversal normalization stays a deliberate, reviewed step.

**Rebase:** the branch had fallen 18 commits behind `main` (the #154/#155/#156 trio, #163, etc.) and gone conflicting. Merged `main` in and resolved the union in `prompts/mcp-base.md` (the #163 instruction-field de-dup plus the require-root PATHS line this PR adds) and in `src/mcp/mod.rs` (kept both the require-root content-assertion and the #163 de-dup assertions; `SERVER_INSTRUCTIONS` byte-lock now 1327). `cargo test` (561 passing) and `cargo fmt --check` green.

The `write.rs` scope-vs-cwd reconciliation stays on the #158 rebase, per your merge note.

---

jahala:
**Reviewed by Opus 4.8:**

# Review — PR #159: require an absolute path/root on read-side file-I/O tools

**Verdict: needs a design decision before merge.** The implementation is clean and well-tested, but the chosen behavior — *hard-refuse* any relative path/scope without an absolute `root` — breaks the common, correctly-configured case and is internally inconsistent with the tool's own documented example.

## What I verified
- `cargo fmt --check`, `cargo clippy -- -D warnings` (exactly what CI runs), and the full test suite all pass on this branch.
- This is the **bottom of a 3-PR stack**: #159 ⊂ #158 ⊂ #157. #159 is the only one that is genuinely read-side-only; #158 and #157 build on top of it. Merge bottom-up (159 → 158 → 157).

## Main concern — the hard-refuse breaks the common case

The server already goes to real lengths to resolve the correct project root at startup (`src/mcp/mod.rs` `run()`): explicit `--scope` > MCP `roots/list` protocol > `package_root(cwd)` > cwd, and it `chdir`s there. So in the normal setup, the process cwd *is* the right checkout, and a bare `tilth_search(query: "foo")` worked correctly.

After this change, that same bare call returns an error (`relative scope … root`) even when cwd is exactly right — every read/search/list/grok/deps call must now carry an absolute `root`, and correctness depends on the model remembering to supply it on every call. That is a heavy reliance on model discipline, and weaker models (haiku — which this project explicitly targets) are the least reliable at it.

**Internal inconsistency that demonstrates the problem:** the prompt gains a "PATHS: pass an ABSOLUTE path/scope, or set `root`…" line, but `AGENTS.md`'s own canonical example is left unchanged:

```
Usage: tilth_search(query: "handleRequest").
```

That example now **errors** under this PR. The docs and the behavior contradict each other.

**The asymmetry is arguably backwards.** The stack's stated rationale is "a wrong-cwd read returns harmless wrong content, so reads can be strict; refusing every relative write would break the in-tree path, so writes warn." But that applies the breaking refusal to *reads* — the most frequent operation and the one where a wrong result is *harmless* — while the genuinely dangerous operation (writes) gets the gentle warn-and-proceed. If a wrong read is harmless, reads are exactly where warn-and-proceed is safe; the hard error buys little and costs the common path.

**Suggested direction:** keep the cwd/startup-root fallback for the common case, and instead *detect the worktree mismatch and warn* (mirroring the write side's `cross_worktree_warning`), rather than refusing outright. If the hard-refuse is intentional, then at minimum (a) update the `AGENTS.md`/prompt examples to all carry `root` so the docs are self-consistent, and (b) run the haiku benchmarks (`rg_search_dispatch`, `rg_trait_implementors`, `gin_servehttp_flow`) — this is exactly the class of instruction/behavior change the project flags as haiku-sensitive.

## Minor / notes (non-blocking)

- **[MINOR] `src/mcp/tools/mod.rs` — `resolve_scope` missing-dir fallback returns a non-canonicalized root.** The explicit-existing-dir branch returns the canonicalized path; the missing-dir fallback returns the raw `root`. Cosmetic (downstream walkers tolerate non-canonical paths), but worth aligning so output path prefixes are consistent.
- **[NOTE] behavior-change trap: an absolute-but-nonexistent `scope` now hard-errors** (when no `root` is given) where it previously warned and fell back to cwd. Intended under the worktree-safety thesis, but existing callers that passed a stale/typo'd absolute scope and relied on the soft fallback will now get a hard error.
- **[NOTE] reads anchor but do not contain.** `root="/repo"` + `path="../../etc/passwd"` resolves and reads outside the repo. This is **not a regression** (pre-PR reads used the raw path and absolute reads were always unconstrained), and `anchor_path`'s non-normalization of `..` is deliberately pinned by a test — just flagging that the read side has no containment guard, unlike writes.

## Things that are solid
- Error propagation is correct: `?` threads the refusal out of every tool (`search`/`read`/`files`/`deps`/`grok`).
- The fallible `resolve_scope` / `anchor_path` / `resolve_read_path` split is clean, and the inverted "no-root → refuse" tests are good regression guards.
- Concurrency: dispatch is sequential, no shared-state hazard introduced. No panics in production paths (all new `unwrap`/indexing is test-only).

paulnsorensen:
_Opus here, Paul had me write this after researching our original bug detection where we discovered this._

**Real-world hit on this exact footgun — argues for refuse over warn-and-fallback.**

We ran into this in a live session, and it's the reason I'd push back on the "keep the cwd fallback, just warn" direction.

What happened: the model `cd`'d into a git worktree partway through the session and then issued a batch of relative-path operations. The MCP server's cwd is frozen at spawn (`run()` chdir's once at startup, plus the optional `roots/list` round-trip — never again), and it's a separate process from the harness shell, so the `cd` was completely invisible to it. Every relative path resolved against the _original_ checkout. Because a worktree mirrors the main tree's layout, the paths all existed there too — so nothing errored. The operations "succeeded," and we had no signal about which tree they actually landed in until we went looking.

That's the core problem with resolving a relative path against a frozen cwd: after the model moves, the path is _fundamentally unknowable_ to the server — it can't observe the `cd`, so it can't tell the intended tree from the stale one. There are only three options:

1. **Refuse** a relative path without an absolute `root` — this PR. The ambiguity can't arise.
2. **Warn and resolve against frozen cwd anyway** — the model sails past the warning, and on the write side you get silent edits to the wrong checkout. A warning on an operation that _looks_ like it succeeded is not a safety mechanism.
3. **Silently resolve against frozen cwd** — the status quo this PR fixes, i.e. what bit us.

The review's asymmetry argument ("a wrong read returns harmless content, so reads can be lenient") actually cuts toward refuse once you follow it onto the write side: a wrong _write_ is not harmless, and the same relative-without-root ambiguity drives both. Making reads strict keeps the model in one consistent discipline — pass an absolute `root` (or absolute path) — rather than "lenient here, strict there," which is exactly the inconsistency that produces a confident wrong call.

Re the ergonomics/haiku concern: agreed it's worth measuring, and the benchmark ask is fair. But the failure mode it's weighed against isn't "a slightly noisier prompt" — it's "writes landed in an unknown tree and reported success." That asymmetry is why I'd keep the hard refuse.

PR #168, opened 2026-07-02
feat(mcp): add `readOnlyHint` annotations to tool definitions

Wedvich:
Adds `readOnlyHint: true` to read-only tools, to allow them to be used in Plan mode in Claude Code.

This used to work just fine for me, but I noticed today that I repeatedly get asked for permission to run the `tilth_read`/`tilth_search` tools, even though I've explicitly added an `"allow"` permission for them in my CC settings. And even if I say "yes, and don't ask again" it keeps asking again. I believe it's caused by this fix in Claude Code 2.1.198:

> https://code.claude.com/docs/en/changelog#2-1-198
> Fixed plan mode not auto-allowing read-only tool calls when a session starts in plan mode

So while it's fixed for tools it _knows_ are read-only  - built-in reads, and a hard-coded set of read-only bash commands - it can't tell `tilth_search` (read) from `tilth_write` (write) unless the MCP server annotates each tool as read-only.

Verification:
- All checks pass locally.
- I replaced my local `cargo install tilth` with the built version of this PR, and the read-only tools now work in Plan mode again without asking for permission.

I didn't add tests for this since it's just an annotation - let me know if you want assertions for it.

`readOnlyHint` reference:
https://blog.modelcontextprotocol.io/posts/2026-03-16-tool-annotations/#what-tool-annotations-are

### DIFF
--- a/src/mcp/tools/definitions.rs
+++ b/src/mcp/tools/definitions.rs
@@ -20,6 +20,7 @@ pub(in crate::mcp) fn tool_definitions(edit_mode: bool) -> Vec<Value> {
     let mut tools = vec![
         serde_json::json!({
             "name": "tilth_search",
+            "annotations": { "readOnlyHint": true },
             "description": "Search for symbols, text, or regex patterns in code. Replaces grep/rg and the host Grep tool — use this for all code search. Symbol search returns definitions first (via tree-sitter AST), then usages, with full source code inlined for top matches. Content search finds literal text. Regex search supports full regex patterns. For cross-file tracing, pass comma-separated symbol names (max 5).",
             "inputSchema": {
                 "type": "object",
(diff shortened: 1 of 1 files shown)


jahala:
Evidence packet for the merge: Claude Code ≥2.1.198 requires the explicit `readOnlyHint` to auto-allow MCP tool calls in plan/auto-accept modes — without it every tilth read re-prompts, which defeats the tool's purpose in exactly the sessions it serves. The annotations mirror tilth's real read/write split (7 read tools true, tilth_write false) and match the MCP spec. Diff reviewed line-by-line; CI green on the exact head. A byte-lock test for the annotation surface lands maintainer-side in a follow-up. Thanks for the contribution — first-time PRs of this shape (small, truthful, spec-grounded) are the easiest yes.

PR #169, opened 2026-07-02
feat: structural scout platform + search determinism fixes (clean port onto main)

jahala:
## What

A clean port of the validated scout line onto main — three commits, rebuilt from main tip rather than merged from the integration workbench:

- **Scout platform** (`048e8e3`): optional local-inference layer (`src/infer/`, Cargo feature `infer`, zero default-build weight), `tilth scout` CLI (`--job context|rerank`, `--warm`, `--hook`), reactive `tilth_scout` MCP tool, `tilth model pull`, trimmed locate core (the footer/git-fusion experiments are deliberately left behind), `TILTH_NO_*` bisect kill-switches, determinism fixes (deterministic `tilth_files`, sorted blast radii, no volatile bytes in MCP results), size-independent embed index (`.gitignore`-aware walker, on-demand pool embedding, warm-only union, format-versioned cache), CoreML compiled-model cache (fixes the `$TMPDIR` leak).
- **Search determinism + honest disclosure** (`d67fc1b`): removes the racy early-quit in all three parallel search walkers — identical searches previously returned *different match sets* run-to-run, and capped regex/content results could silently read as complete. Walks now complete, results are byte-deterministic, headers disclose "N of M matches".
- **Instruction lines** (`9127759`): restores the two measured steering lines (grok for call-chains; batching for multi-symbol/multi-section calls) with updated byte-lock tests.

## Receipts

- **A/B (main tip vs this branch, Haiku, 15 tasks × 3 reps/arm): −3.1% cost-per-correct, accuracy identical 45/45** — verdict LAND (do-no-harm gate for the platform; the win is lineage hygiene + determinism).
- Scout deep-trace A/B (150 runs): hook payload delivery 75/75 correct anchors, 0 stray fires, accuracy unchanged — the tool is safe; the **hook remains opt-in** (`tilth install --with-scout`) pending the re-aimed value test.
- Gates: full test suite green, `clippy -D warnings` clean, `fmt --check` clean; MCP fire-map verified on the release `--features infer` build.

## Not in this PR

No version bump (release PR comes separately per the release workflow). No footer (measured net-negative; not ported). No hook-by-default. Follow-up branches stack on this and arrive as separate PRs: grok qualified-name resolver, language-coverage fixes (Ruby/C/C++/Go), response-format v2.


paulnsorensen:
> [!NOTE]
> **Automated review** — posted by a Claude Code review agent on behalf of @paulnsorensen. Treat findings as advisory; a human hasn't vetted each one.

Reviewed across correctness/security/encapsulation/spec/complexity/deslop/assertions/NIH/efficiency/telemetry. The determinism work is solid and well-tested; the blocker is merge-staleness, not the change itself.

**Blocker — will not compile after merge into current `main`**
`src/mcp/tools/scout.rs` calls `resolve_scope(args)` (1-arg, returns a tuple) — the signature from this branch's base. `main` has since replaced it (`ed51e58`) with `resolve_scope(args, root) -> Result<(PathBuf, Option<String>), String>`. The `let (scope, scope_warning) = resolve_scope(args);` destructure fails on both arity and Result-vs-tuple. CI here passed against the older base (2026-07-02); `main` moved. Rebase and port scout's scope resolution to the 2-arg `Result` form (threading an absolute `root` and obeying the require-root discipline like the other tools).

**Medium**
- `scout.rs` prepends the plain-text `resolve_scope` warning in front of the JSON payload the tool schema documents as "Returns JSON". On an invalid/relative scope the result is `<prose>\n\n{json}` and won't `JSON.parse`. Put the warning inside the JSON object instead.
- `install.rs::write_claude_code_scout_hook` writes the user's `~/.claude/settings.json` with a non-atomic `fs::write`; an interrupt corrupts the host's central settings. This file already has `atomic_write` — use it.

**Low**
- `search/symbol.rs::find_usages`: the `// Early termination: enough usages found` comment survives above the deleted quit block — remove it.
- `search/symbol.rs`: after the quit removal, `found_count` (`fetch_add` in `find_definitions`/`find_usages`) is write-only; drop it.
- `search/content.rs` + `symbol.rs`: removing the walk-time cap makes match accumulation unbounded in total-match count. The determinism fix is right; consider a generous hard ceiling for huge-repo/common-token inputs.
- `model_pull.rs`: remote pull has no checksum and accepts `http://` for a natively-executed ONNX file. Prefer https-only or a pinned digest. (Opt-in feature + user-supplied URL, so low.)

**Coordination**: `search/glob.rs` is rewritten here and is also touched by #185 — these will collide in `glob::search`; reconcile on merge order.

PR #177, opened 2026-07-05
feat(list): fold tilth_files into tilth_list with directory-tree rendering

paulnsorensen:
Folds the `tilth_files` tool into `tilth_list`: `patterns` becomes required, adds a `depth` parameter, and adds token-rollup directory-tree rendering (`mcp/tree.rs`). Prompts, AGENTS.md, and the byte-lock tests are updated in lockstep.

Part of the fork-divergence port (G3). Reviewed via fresh-context taste-test; `cargo test` (622), clippy `-D warnings`, and fmt all green.

## Size sample

Same `tools/call` — `{"patterns": ["*.rs"], "scope": …}` over the identical checkout — sent over MCP stdio through debug builds of main (`5fa0c86`, flat `tilth_files`) and this branch (`tilth_list` tree). Sizes are UTF-8 bytes of the returned content text.

| scope | old flat (main) | new tree (branch) |
|---|---|---|
| `src/` — 68 files, nested | 715 B, capped at 20/68 files + "narrow with scope" hint | 3,394 B, complete: all 68 files with per-dir token rollups |
| `src/search/` — 18 files, flat | 566 B, complete | 726 B, complete (+160 B, +28%) |

Read: on a flat under-cap directory the tree costs ~28% more (connectors + root rollup). On a nested scope the flat view is smaller only because it truncates at 20 files; the tree delivers the complete 68-file structure in one call, where the flat view leaves the agent to re-issue narrowed calls for the remaining 48.


jahala:
Reviewed by Fable 5:

Strong port, and the lockstep discipline (prompts + AGENTS.md + byte-locks in one commit) is exactly right. Verified on this side: the server-side missing-`patterns` error is recoverable and names the fix; the collapsed `pattern`/`patterns` schema is a net tightening; the tree renderer is stateless (clean on the concurrency/state lens); based on current main; the fold lock-test is a nice touch.

Two things before merge:

1. **A silent dispatch alias for `tilth_files` for one release.** Hosts cache the tool list at connect, so any live session spanning the upgrade will call the old name and hard-fail mid-flight. Accept `tilth_files` at dispatch (route to the same handler, not advertised in tools/list — zero schema cost), drop it next minor. ~10 lines including the test.
2. **A before/after size sample** for one representative listing (old flat list vs tree+rollups on the same scope) in the PR body. A name-swap with identical DO-NOT steering semantics doesn't need the full benchmark gate (reasoned exemption), but output-shape deltas belong on the record — we've been burned measuring these by intuition.

With those, this merges.

paulnsorensen:
Review sweep applied (fresh `/age` pass — no outstanding human/bot comments on this PR). Glob-matching semantics verified byte-identical to the old `tilth_files` path.

Changes in `4065659`:

- **Efficiency** — `depth` is now pushed into the walker as `.max_depth(d)` so the walk itself is bounded, instead of walking the whole tree then filtering (off-by-one preserved: top-level file kept at `depth:1`).
- **Float depth** — `depth: 2.0` was silently dropped by `as_u64`; added an `as_f64` fallback so float-encoded JSON numbers are honored.
- **Formatter** — `fmt_tokens` gained an `M` scale and now rounds instead of truncating, matching `map.rs` (minor intended output change, e.g. `~4.3k` vs old `~4.2k`).
- **No-match UX** — restored the "found extensions: …" hint the old glob path emitted on zero matches.
- **Tests** — added two that can actually fail: dir-rollup-equals-sum-of-children, and truncation-note-under-budget.
- **Docs** — `benchmark/README.md`: `tilth_files/run` → `tilth_list/run`.

Plus `c2a31e3`: fixed a pre-existing `clippy::question_mark` lint in `find_git_root` (`write.rs`) that the CI toolchain (rust 1.97) newly enforces — behavior-identical `?` rewrite.

Gates: `cargo test` (612 passed), `clippy -D warnings` (verified on 1.97), `fmt` — all clean.

paulnsorensen:
> [!NOTE]
> **Automated review** — posted by a Claude Code review agent on behalf of @paulnsorensen. Treat findings as advisory; a human hasn't vetted each one.

## Review: fold tilth_files → tilth_list

Consolidation is correct and the prompts / AGENTS.md / byte-lock are in lockstep (CI green, 622 tests). A few cleanups:

**Medium — duplicated token formatter**
`src/mcp/tree.rs` `fmt_tokens` is a verbatim copy of `src/map.rs::fmt_tokens`, including the subtle `999_950` k→M rollover threshold and its `clippy::cast_precision_loss` allow. The doc comment even acknowledges the coupling. map.rs has a dedicated `fmt_tokens_rolls_over_at_999_950` regression test; the tree.rs copy has **no** rollover coverage, so a future change to the threshold in one file diverges silently in the untested copy. Suggest extracting the tier/rounding logic into one shared helper (e.g. next to `estimate_tokens` in `types.rs`), parameterizing only the suffix (`"12.3k"` vs `"~12.3k tokens"`).

**Low — stale comment**
`src/mcp/tools/mod.rs:168` still references the removed tool: `// ... a bare tilth_search/tilth_files call with no scope`. Rename to `tilth_list`.

**Low — misleading test name/assertion**
`render_tree_dir_rollup_equals_sum_of_child_tokens` asserts the dir rollup equals the *sum of child token estimates*, but the code computes `estimate_tokens(total_bytes)`. Since `estimate_tokens` ceil-divides by 4, these diverge for byte counts not divisible by 4 (two 101-byte files → children 26+26=52 vs dir 51); the divisible fixture masks it. Rename/retarget the assertion or sum child estimates.

**Low — weak depth test**
`tool_list_depth_caps_nesting` only asserts the top-level `Cargo.toml` is present; it never asserts a deeper file is excluded, so it can't fail if `depth` were a no-op. Add a negative assertion that `src/*.rs` is absent at `depth: 1`.

jahala (commented):
Reviewed by Fable 5:

The depth-bounding and float-parsing fixes look right. It appears the review sweep missed my earlier feedback (it was posted as an issue comment — this one goes in as a review comment so tooling sees it): two items remain open before merge. **(1)** A silent dispatch alias for `tilth_files` for one release — hosts cache the tool list at connect, so live sessions spanning the upgrade will call the old name and hard-fail; accept it at dispatch (same handler, not advertised in tools/list — zero schema cost), drop next minor. **(2)** A before/after size sample for one representative listing (flat vs tree+rollups) in the PR body. Both small; with them this merges.

PR #178, opened 2026-07-05
refactor(lang): consolidate per-language data into LangSpec table

paulnsorensen:
Consolidates per-language dispatch and data into a LangSpec table: `src/lang/spec.rs` plus 19 per-language modules, with a `NodeTextMode` parameter and all consumer seams converted. Zero behavior change — verified against the old match arms.

Part of the fork-divergence port (G1). Reviewed via fresh-context taste-test; `cargo test`, clippy `-D warnings`, and fmt all green.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -873,6 +873,7 @@ dependencies = [
  "tree-sitter-java",
  "tree-sitter-javascript",
  "tree-sitter-kotlin-ng",
+ "tree-sitter-language",
  "tree-sitter-md",
  "tree-sitter-php",
  "tree-sitter-python",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ clap_complete = "4"
 
 # Tree-sitter (AST outlines)
 tree-sitter = "0.26"
+tree-sitter-language = "0.1"
 streaming-iterator = "0.1"
 tree-sitter-rust = "0.24"
 tree-sitter-javascript = "0.25"
--- a/src/lang/bash.rs
+++ b/src/lang/bash.rs
@@ -1,0 +1,21 @@
+//! Bash language spec.
+
+use crate::lang::spec::{LangSpec, StdlibRule, StripFamily, DEFAULT_DEFS, DEFAULT_DEF_KINDS};
+
+const CALLEE_QUERY: &str = "(command name: (command_name) @callee)\n";
+
+pub(crate) const SPEC: LangSpec = LangSpec {
+    display: "Bash",
+    extensions: &["sh", "bash", "bats"],
+    filenames: &[".bashrc", ".bash_profile", ".bash_aliases", ".profile"],
+    grammar: Some(tree_sitter_bash::LANGUAGE),
+    callee_query: Some(CALLEE_QUERY),
+    sibling_query: None,
+    stdlib: StdlibRule::None,
+    manifests: &[],
+    definition_kinds: DEFAULT_DEF_KINDS,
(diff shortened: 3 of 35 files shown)


jahala:
Reviewed by Fable 5:

Right structure, and the zero-behavior-change claim held up under adversarial reading: the table is all-`&'static` (no interior mutability — clean on the state/race lens), `DEFAULT_DEF_KINDS` preserves the global list, and ruby.rs faithfully carries only its callee query — the refactor honestly preserves even the known gaps, which is exactly what a pure refactor should do.

That last point is why this PR is load-bearing: we have staged fixes for real definition-blindness in Ruby (bare `method`/`class`/`module` kinds missing entirely), C/C++ (function names extracted as anonymous via nested declarators), and Go (`type_declaration` nesting) — currently expressed against the old flat lists. Your table is their right permanent home: per-language `definition_kinds` in each module instead of edits to four shared surfaces. So **this PR sequences first**; the fixes follow as spec-table entries on top, with a 22-language behavioral harness as the acceptance suite. Expect that follow-up shortly after this merges.

Two small things:

1. Confirm the recently-added direct tests (the `definition_weight` tier assertions and `extract_definition_name` cases) are among your 622 — CI suggests yes; just want it stated since they're the behavioral pins for exactly the surfaces this moves.
2. Noted, not blocking: `callee_query.rs:48` compiles the tree-sitter `Query` per call — pre-existing (it arrived with the Bash-support work), but with per-language query strings now consolidated in one table, a once-per-language cache became easy if you're inclined. Happy to take it as a follow-up otherwise.

paulnsorensen:
Review sweep applied (fresh `/age` pass — no outstanding human/bot comments on this PR). Consolidation verified field-by-field against the deleted per-language sources: zero drops or swaps.

Changes in `bcbae22`:

- **Bash dotfiles no longer reclassified** — the consolidation started routing through `detect_file_type`, which (unlike the old `path.extension()?`-first logic) classified extension-less bash dotfiles (`.bashrc`, `.profile`, …) as `Code(Bash)` and stripped them. Gated the route to extension-bearing paths so the "zero behavior change" claim holds. Added regression test `extensionless_bash_dotfile_not_stripped`.
- **Deleted redundant wrapper** — `default_defs_extract_name` was a pure forwarder; wired `default_extract_definition_name` directly.
- **O(1) extension lookup** — replaced the per-file linear scan over `ALL_LANGS` with a static `EXT_TO_LANG` map built from the same `LangSpec` extension data (first-wins tiebreak preserved).
- **Removed intra-crate back-dependency** — hoisted `with_query` into `treesitter.rs` so `lang` no longer depends on `search::siblings`.

Plus `cd09698`: fixed a pre-existing `clippy::question_mark` lint in `find_git_root` (`write.rs`) that the CI toolchain (rust 1.97) newly enforces — behavior-identical `?` rewrite.

Gates: `cargo test` (622 passed), `clippy -D warnings` (verified on 1.97), `fmt` — all clean.

jahala (commented):
Reviewed by Fable 5:

Both review items delivered — the direct tests are confirmed green on an independent run, and the query cache is better than asked: entry-atomic insertion and poison-recovery via `into_inner` avoid two failure modes we've measured elsewhere in this codebase. One optional nit, not blocking: `with_callee_query` runs the callback under the global cache mutex, so callee-query executions serialize process-wide — storing `Arc<Query>` per entry and cloning out before invoking would unserialize it, whenever you feel like it. Merge-ready from this side; sequencing stands (this lands ahead of the language-coverage follow-up described earlier).

Issue #16, opened 2026-03-06
Consider documenting swift support


evanmoran:
I love this. Thank you for putting it together! I'm passing it around to everyone I know, as it's insanely good :) 

I was about to create a pull request to add swift support, when to my surprise it already works. Consider indicating in the readme that it works (and other languages too if they also work). 

Thank you again! 

-Evan

evanmoran:
You are quite welcome! It's well deserved as this tool is super useful.  And thank you for adding support so quickly!

PR #180, opened 2026-07-09
fix: 15 bugs from whole-src bug hunt (correctness, security, memory, contract)

paulnsorensen:
Whole-`src/` bug hunt (no diff in scope) run as a 10-dimension `/age` fan-out, reconciled and re-verified against source. **15 of 17 selected findings fixed** across 6 file-disjoint commits; 2 deferred to follow-ups (below). A fresh-context reviewer re-derived every behavioral fix against source and returned **PASS**.

## Gates
- `cargo fmt --check` ✓ · `cargo clippy --all-targets -- -D warnings` ✓ · `cargo test` **835 passed, 0 failed**
- Version untouched at 0.8.4 (fork law).

## Fixed

**Blockers**
- `edit/apply.rs` — `append`/`Cursor::Tail` inserted a spurious blank line and dropped the trailing newline on every append to a newline-terminated file. Now resolves before the phantom split-row.
- `prompts/mcp-base.md` — shipped instructions told agents `tilth_write` takes `edits: "..."` (a string) while the code requires a JSON array. Corrected on all three surfaces (prompt, `AGENTS.md`, schema); byte-locks updated.

**High**
- `diff/mod.rs` — git argument injection: agent-controlled `source`/`log` reached `git diff`/`git log` unguarded (`--output=…` overwrites arbitrary files outside cwd confinement). Leading-dash guard at both sinks.
- `edit/apply.rs` — two zero-width inserts at the same index applied in reversed author order.
- `mcp/tools/write.rs` — `move_file` silently clobbered an existing destination. Now rejected.
- `prompts/mcp-edit.md` — instructions claimed `tilth_search` emits a `[path#TAG]` header (it never does). Removed.
- `cache.rs` + `index/bloom.rs` — unbounded server-lifetime caches; now `clru`-bounded (mirrors `edit/snapshots.rs`), bloom sized on unique idents.

**Medium**
- `diff/parse.rs` — hunk-internal lines starting `-- `/`++ ` (e.g. deleted SQL comments) were misparsed as file headers and dropped. Header branches now gated to outside-hunk.
- `mcp/iso.rs` — `if_modified_since` silently rejected fractional-second/offset ISO timestamps (JS `toISOString()`), disabling cache redaction. Now accepted.
- `mcp/mod.rs` + `grok.rs` — `tilth_grok` validated a `budget` param it ignored; now honored + advertised.
- `mcp/tools/list.rs` — invalid globs silently dropped; now rejected loudly.
- `edit/snapshots.rs` + `session.rs` — snapshot-key normalization centralized (`impl AsRef<Path>`), removing caller-shadowed-invariant risk.
- `mcp/tools/read.rs` — fuzzy resolution used the process cwd instead of the per-call `cwd`.
- `install.rs` — hand-rolled TOML writer (no quote escaping) + lexer replaced with `toml_edit`.
- `search/symbol.rs` + `content.rs` — duplicated file-walk preamble extracted to `accept_walk_entry`.

## Deferred (flagged, not dropped — each warrants its own PR)
- **#16** `search/mod.rs` god-module refactor — a 3.2k-line-file, zero-behavior-change extraction; too large to fold into a bug-fix PR.
- **#8** `.tilthignore` producer-move (security, high) — investigation showed the deny must cover multiple independent read paths (incl. `git diff` shelling out, which no Rust fs-guard covers, and a cache path that bypasses `read_file`). A partial guard would falsely signal "secrets safe everywhere". Needs a dedicated, verified PR — recommend scheduling first.

Full report: `.cheese/age/src-bug-hunt.md` · cure log: `.cheese/cure/src-bug-hunt.md`.


paulnsorensen:
Opened against the wrong repo — this belongs on the fork. Recreated as https://github.com/paulnsorensen/tilth/pull/126.

PR #181, opened 2026-07-11
ci: lint test targets with clippy --all-targets

paulnsorensen:
## Why

The `check` job ran `cargo clippy -- -D warnings`, which never compiles `#[cfg(test)]` code. The crate sets `#![warn(clippy::pedantic)]`, but that bar only reached test modules when a violation happened to also live in lib code — so pedantic lints in tests slipped through PRs and could only ever break `main`.

This is exactly what just happened: the recent clippy bump surfaced a lint that had accumulated uncaught.

## What

- **CI**: `cargo clippy -- -D warnings` → `cargo clippy --all-targets -- -D warnings`, so test/bench targets are linted at PR time.
- **Clear the 61 pre-existing test-target violations** the flag exposes:
  - `format_push_string` → `writeln!`
  - `format_collect` → `fold(String::new(), …)` with `writeln!`
  - `needless_raw_string_hashes`, `single_char_pattern`, `redundant_closure_for_method_calls`, `doc_markdown`, `unnecessary_to_owned` → `clippy --fix` (all in test modules / doc comments)
- Update the `CLAUDE.md` build + CI notes to match.

All changes are confined to `#[cfg(test)]` modules and their doc comments — no production-path behavior changes.

## Verification

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` (clean), `cargo test` (608 passed) all green locally.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -7,5 +7,5 @@
 <!-- How to verify this works. CI checks fmt + clippy + test on every push. -->
 
 - [ ] `cargo fmt --check`
-- [ ] `cargo clippy -- -D warnings`
+- [ ] `cargo clippy --all-targets -- -D warnings`
 - [ ] `cargo test`
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
           components: clippy, rustfmt
       - uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # ratchet:Swatinem/rust-cache@v2
       - run: cargo fmt --check
-      - run: cargo clippy -- -D warnings
+      - run: cargo clippy --all-targets -- -D warnings
       - run: cargo test
 
   markdown:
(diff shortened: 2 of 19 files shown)


jahala (commented):
Reviewed by Fable 5:

Verified to an unusual depth, and it holds: zero assertion lines changed (grep-proven both directions), all 72 hunks inside test-only regions, every rewritten fixture generator byte-identical to its original (checked mechanically, including the large OGATE fixtures), six threshold-sensitive tests mutation-tested — all six still catch their mutants — and zero `#[allow]`s added: 61 real fixes, no silenced debt. The 61 count is exact on current stable; the earlier differing counts (63/65) were toolchain skew, which your PR's premise survives either way — the old CI invocation sees none of them. Stale CI is covered: we re-ran all gates on the post-merge tree against today's lockfile, clean.

Two one-line asks, then this merges: `CONTRIBUTING.md:11` and `.github/PULL_REQUEST_TEMPLATE.md:10` still instruct the old `cargo clippy -- -D warnings` — after this lands, that command passes locally while CI fails, precisely the divergence you're eliminating. Both become `--all-targets`.

Courtesy notes, no action needed: the body says 608 tests (it's 609), and lists `unnecessary_to_owned` among the fixes (the actual pair was `useless_format`/`needless_borrow`).

PR #184, opened 2026-07-14
fix(search): skip .jj/.hg/.svn VCS internal dirs

yasyf:
## Summary

`SKIP_DIRS` skips `.git` but not the other VCS internal directories.
In a jj-colocated repo (`jj` over git), the entire `.jj` object store —
easily thousands of files — gets walked on every search, outline, and
glob. `.hg` (Mercurial) and `.svn` (Subversion) are the same class of
internal store. This adds all three next to `.git`; they're VCS
internals, never source, exactly as the list's own comment frames it.

Repro (before this change):

    mkdir /tmp/jjrepo && cd /tmp/jjrepo
    git init -q && jj git init --colocate -q   # creates a large .jj/ store
    echo 'fn main() {}' > main.rs
    tilth "*" --scope .                        # .jj/**/* is walked and counted

After: the `.jj` tree is skipped, the same as `.git`.

## Test plan

Adds `walker_skips_vcs_internal_dirs`: a tempdir with `.jj`/`.hg`/`.svn`
subdirs, each holding a source file, asserts the walker returns the
top-level source and none of the buried files.

- [x] `cargo fmt --check`
- [x] `cargo test`

### DIFF
--- a/src/search/mod.rs
+++ b/src/search/mod.rs
@@ -39,6 +39,9 @@ use crate::format::rel;
 // gitignored files (docs/, configs, generated code) are still searchable.
 pub(crate) const SKIP_DIRS: &[&str] = &[
     ".git",
+    ".jj",
+    ".hg",
+    ".svn",
     "node_modules",
     "target",
     "dist",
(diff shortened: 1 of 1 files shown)


paulnsorensen (approved):

jahala (commented):
Reviewed by Fable 5:

Verified red-on-main (VCS-internal files demonstrably leak through the walker without this) and the fix lands in exactly the right place — the shared `SKIP_DIRS` feeding every walker via `base_walk_builder`, no second list; the basename-fallback walker is already covered by its hidden-dir exclusion. Nothing added beyond the fix. Merging once the identity item below is sorted.

One housekeeping ask that applies to all four of your PRs: the commits are authored as `contributor <contributor@example.com>` — a placeholder. Amend with a real identity (or we'll squash preserving your GitHub authorship) so the credit lands where it belongs in `git log`.

PR #185, opened 2026-07-14
fix(search): make glob file selection deterministic

yasyf:
## Summary

Glob matches are pushed into a `Mutex<Vec>` from `WalkParallel` under an
`if len < MAX_FILES` cap. Because the parallel walk visits files in a
nondeterministic order, two identical glob runs return different 20-file
subsets of the same match set. This collects every match, sorts by path,
then caps — the returned files are now the lexicographically smallest
matches in sorted order, stable across runs.

Output shape is unchanged beyond that determinism: `total_found` still
counts every match, the "… and N more files" tail is unchanged, and
previews are now computed only for the surviving files instead of for
every match before the cap.

Repro (before this change):

    cd a-repo-with-more-than-20-matches
    tilth '*.rs' > a.txt
    tilth '*.rs' > b.txt
    diff a.txt b.txt   # differs run to run

Overlaps with #186 in `glob.rs`; happy to rebase whichever lands second.

## Test plan

- `glob_selection_is_deterministic_and_sorted`: 35 matching files, asserts
  two runs return the identical set and that it's the sorted first-20.
- `glob_total_found_counts_every_match`: asserts the list stays capped at
  `MAX_FILES` while `total_found` reports the full count.

- [x] `cargo fmt --check`
- [x] `cargo test`

### DIFF
--- a/src/search/glob.rs
+++ b/src/search/glob.rs
@@ -29,16 +29,14 @@ pub fn search(pattern: &str, scope: &Path) -> Result<GlobResult, TilthError> {
     })?;
     let matcher = glob.compile_matcher();
 
-    let files: std::sync::Mutex<Vec<GlobFileEntry>> = std::sync::Mutex::new(Vec::new());
-    let total_found = std::sync::atomic::AtomicUsize::new(0);
+    let matched: std::sync::Mutex<Vec<PathBuf>> = std::sync::Mutex::new(Vec::new());
     let extensions: std::sync::Mutex<HashSet<String>> = std::sync::Mutex::new(HashSet::new());
 
     let walker = super::walker(scope, None)?;
 
     walker.run(|| {
         let matcher = &matcher;
-        let files = &files;
-        let total_found = &total_found;
+        let matched = &matched;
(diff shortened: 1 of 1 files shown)


paulnsorensen (approved):

jahala (commented):
Reviewed by Fable 5:

Confirmed the bug and the fix at CLI level: 12 identical runs on main returned 5 different 20-file sets; your branch returns one, totally ordered (`PathBuf: Ord` over unique paths is a complete order — no tiebreak needed for file selection), with `total_found` captured before the truncate so its semantics are preserved. Measured cost on a 60k-match tree: ~+0.05s wall, ~+2MB — acceptable, and offset by previews now running on 20 files instead of all matches. Also checked the selection-quality worry: main clusters into one directory just as hard as sorted order does — your change only makes it reproducible.

Merging — and deliberately sequenced *ahead* of #186: behind this cap its preview pass costs nothing; without it, every matched file gets opened (we measured 3–6× on a 20k-match glob). Same commit-identity ask as on #184.

PR #186, opened 2026-07-14
fix(search): show size + type for binary files in glob previews

yasyf:
## Summary

Glob file previews run `estimate_tokens` (bytes / 4) over fs metadata for
every file, so a PNG, WASM blob, or font lists a large, meaningless token
estimate — a binary is not N tokens of text. This detects binary content
the same way the read path does (a null byte in the first 512 bytes) and
renders `binary, <size>, <mime>` instead, reusing `format::format_size`
and `read::mime_from_ext` so the phrasing matches `tilth_read`'s binary
header. Text files are unchanged and still show the token estimate.

Repro (before this change):

    tilth '*' --scope dir-with-images
    #   logo.png  (~2.1k tokens)   ← meaningless for a binary

After:

    #   logo.png  (binary, 8KB, image/png)

Overlaps with #185 in `glob.rs`; happy to rebase whichever lands second.

## Test plan

Adds `file_preview_renders_binary_type_not_tokens`: a `.png` with a null
byte previews as `binary, 12B, image/png`; a `.rs` still ends in
`tokens`. `format_size` / `mime_from_ext` are promoted to `pub(crate)`.

- [x] `cargo fmt --check`
- [x] `cargo test`


jahala (commented):
Reviewed by Fable 5:

Good reuse discipline — the shared `lang::detection::is_binary` (same 512-byte window as the read path — no second heuristic to diverge later), `format_size`, `mime_from_ext`. The adversarial cases behave: device symlinks and FIFOs never reach the preview (the walker's `is_file` guard filters them), and an 8GB sparse file costs one bounded 512-byte read. The stat-then-read race is cosmetic (a preview string, no safety decision).

Three small asks, then this merges — after #185, for the performance reason noted there:
1. `format!("binary, {size}, {mime}")` duplicates the fragment inside `format::binary_header` — extract one helper, call it from both.
2. `file.read(&mut head)` can short-read; use `file.take(512).read_to_end(..)` so the probe window is guaranteed.
3. Your change surfaces a pre-existing `format_size` gap on a much hotter path: 8GB renders as `8192.0MB` (no GB tier). Add the tier or state it as accepted in the body.

Same commit-identity ask as on #184.

PR #187, opened 2026-07-14
fix(search): handle binary and non-UTF-8 files in grok without a raw decode error

yasyf:
## Summary

`grok` reads its target file with `fs::read_to_string`, so a binary or
otherwise non-UTF-8 code file — a `.rs` with a null byte, or a
`path:line` target pointing at one — surfaces a raw
`stream did not contain valid UTF-8` error instead of a clean message.
Plain read already guards this: it detects binary content and decodes the
rest lossily.

`read_code_file` now mirrors the read path — it checks `is_binary` and
refuses binary files with an actionable message, and decodes the rest
with `from_utf8_lossy`, so a code file with a few invalid bytes still
outlines instead of erroring. The extension check runs first, so a
non-code file keeps its existing "not a code file" message.

Repro (before this change):

    printf 'fn main() {}\n\x00\x01\x02' > tainted.rs
    tilth grok tainted.rs:1
    # grok error: tainted.rs: stream did not contain valid UTF-8

After:

    # grok error: invalid query "tainted.rs": binary file — grok needs source code

## Test plan

- `resolve_by_path_line_rejects_binary_code_file`: a `.rs` with a null
  byte returns `InvalidQuery` (was a raw IoError), reason contains
  "binary".
- `resolve_by_path_line_reads_non_utf8_code_file_lossily`: a `.rs` with
  an invalid byte (no null) resolves via lossy decode.

- [x] `cargo fmt --check`
- [x] `cargo test`


jahala (commented):
Reviewed by Fable 5:

The non-UTF-8 half is a genuine fix — verified: a `\xff` file errors raw on main exactly as described. The binary-refusal half is something *better* than what the PR says it is: null-byte files actually grok fine on main (your own test fails against main because main **succeeds** — null bytes are valid UTF-8), so the refusal is a behavior change, not a bug fix. And it's the right change, for a stronger reason than the body gives: `read_to_string`'s strictness was an accidental DoS guard, and lossy decoding removes it — we measured the lossy-only variant at 6.7s / 1.07GB peak RSS on a 5MB garbage `.rs`; your `is_binary` check is what puts an explicit guard back. Keep it — reframe it.

Asks before merge:
1. **Tell the true story**: split into `fix(grok): decode non-UTF-8 source lossily` + `feat(grok): refuse binary targets` (or rewrite the body/test comment): the refusal is deliberate and closes a parser-DoS, not "surfacing a raw decode error."
2. **Close the probe hole**: 512 clean ASCII bytes followed by megabytes of garbage evades `is_binary` — measured 6.9s / 254MB through your branch where main errors instantly. Bound the lossy path by size (precedent: `full_read_size_cap`) and carry that fixture as a test.
3. **Make the refusal recoverable in one step**: `tilth_search` *does* index these files (verified), so right now search advertises what grok refuses and the agent is stranded. The error should point there: "binary or non-text file — grok needs source; use tilth_search \"<symbol>\" for matches in this file."
4. When the refusal fires via the symbol route (`enrich_from_outline`), the error carries the resolved *path* as the "query" — carry the user's actual query.
5. Preserve file-not-found precedence for non-code extensions (a typo'd `.md` path now reports "not a code file" instead of missing-file).
6. Zero-copy the happy path: `String::from_utf8(bytes)` first, lossy only on failure — the current `from_utf8_lossy(&bytes).into_owned()` doubles peak memory for the ~always-valid case.

The instinct across all four PRs is exactly right — this one just needs its true story told. Same commit-identity ask as on #184.

PR #200, opened 2026-08-16
perf(build): slim dev debug info to line-tables-only

paulnsorensen:
Cuts target-dir size and dev build memory for parallel agent gates; backtraces keep file:line. Matches tern/oh-my-pi. Refs crabbot#46.


paulnsorensen:
Opened in error by an automated fan-out. This dev-profile tweak was only meant for the local checkout on my machine, not a contribution to upstream. Closing; apologies for the noise.

PR #216, opened 2026-09-12
fix(grok): see Rust calls inside macros and count Rust test code as tests

andoriyu:
## Summary

In Rust code, `grok` and `callers` missed most test call sites, and `grok` never showed a tests section; this fixes both causes. First, tree-sitter-rust parses a macro's arguments as an opaque `token_tree`, so `assert!(greet("world"))` contains no `call_expression`: the Rust callee query (`src/lang/rust.rs`) now also matches an identifier followed by a parenthesised token tree inside a token tree, which covers `assert!`, `assert_eq!`, `vec![]`, `format!` and similar macros, for callers and callees alike. Second, `is_test_file` only knows `.test.`, `.spec.` and `__tests__/`, which Rust never uses, so Rust tests were listed as callers: `CallerMatch` now records `in_test` for a file under a `tests` directory (relative to the scope), a function marked `#[test]` or `#[<path>::test]` (such as `#[tokio::test]`), or a `#[cfg(test)]` module, and `grok` counts a match as a test when either signal says so. No tool schema changes.

## Test plan

Regression tests `grok_finds_rust_callers_inside_macro_arguments` and `grok_counts_rust_test_dirs_and_test_functions_as_tests` fail on `main` and pass with this change.

To see it by hand, take a crate with `greet` in `src/lib.rs`, a caller `welcome()`, a `#[cfg(test)]` module, and `tests/greet.rs`, then grok `greet`.

Before (0.10.1): the calls inside `assert!` (`tests/greet.rs:5`) and `assert_eq!` (`:10`) are missing, and there is no tests section.
```
## callers (3)
  src/lib.rs
    [8]   in welcome()
    [17]   in tests.a_unit_test()
  tests/greet.rs
    [15]   in a_hoisted_call()
```

After:
```
## callers (1)
  src/lib.rs
    [8]   in welcome()

## tests (4)
greets_by_name                      tests/greet.rs:5
a_greeting_is_a_sentence            tests/greet.rs:10
a_hoisted_call                      tests/greet.rs:15
tests.a_unit_test                   src/lib.rs:17
```

- [x] `cargo fmt --check`
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `cargo test`

https://claude.ai/code/session_01QvxtApauezMf17e5uqdUkr


jahala:
Reviewed by Fable 5.1

Both halves are real. Full diff read; fmt, clippy and tests green on stable 1.98.1. On tilth's own tree `grok estimate_tokens` now splits 11 tests out of what was 43 callers, which is exactly the section that was missing.

Two things before it can land:

1. The macro-argument half overlaps #218, opened half an hour after yours. It takes the same route through `token_tree` but requires the next tree to open with `(` and filters the `assert_eq!(x, (a, b))` case, so `idx[1]` and `Foo { .. }` are not read as calls. That version is landing for that half, and the two conflict in `src/lang/rust.rs`. Once #218 is in, could you rebase and drop the `CALLEE_QUERY` hunk? Your `grok_finds_rust_callers_inside_macro_arguments` test passes on #218's pattern, so keep it.

2. The `in_test` detection is the part #218 does not have, and it is wanted. `is_rust_test_site` guards on `lang != Lang::Rust` inside language-agnostic `callers.rs`; #218 adds `callee_filter` to `LangSpec` for exactly this kind of per-language hook, so a `test_site` slot beside it would keep `callers.rs` neutral. Not a blocker if you would rather keep the diff small.

andoriyu:
@jahala Updated this after the rebase.

jahala:
Reviewed by Fable 5.1

Thanks for the rebase. This is what the review asked for: the overlapping query hunk is gone, your macro-argument test stays and passes on #218's pattern, and test detection now sits behind a `test_site` slot in `LangSpec`, so `callers.rs` stays language-neutral.

I read the full diff and ran it merged with current main on stable 1.98.1: fmt, clippy `-D warnings` and the test suite are green. On tilth's own tree, `grok estimate_tokens` now reports `tests (8 of 11)` separately from `callers (5 of 32)`, where main lists them all as callers.

One limit worth knowing, not a blocker: attribute detection matches `#[test]`, `#[<path>::test]` and `#[cfg(test)]`. Test macros with other names, such as `#[rstest]` or `#[test_case(..)]`, and compound predicates like `#[cfg(all(test, ..))]`, still count as callers. A follow-up is welcome if you hit those.

Merging.

Issue #215, opened 2026-09-12
tilth_search: expose the `full` result cap over MCP, as tilth_grok already does


plusky:
## Problem

Over MCP, `tilth_search` always runs with `full = false`: `src/mcp/tools/search.rs` passes the literal into all six dispatches (symbol single and multi, content, regex, callers single and multi), so every kind is capped at `MAX_MATCHES = 10` and no argument can raise it. The CLI has had `--full` for this since #111 (cap 100).

Measured on a Rust workspace: a content search for `?Capped(` restricted to one file returned 10 of the 15 sites; a symbol search for `Capped` returned 10 of 69 usages. "Narrow with scope" does not help when the hits are in one file. For a rename or an audit the agent then has to fall back to grep, which the server instructions tell it not to do.

## Proposed shape

- `tilth_search` gains `"full": { "type": "boolean", "default": false }` with the CLI flag's semantics: raise the match cap from 10 to 100 for every kind. `tilth_grok` already exposes an identical boolean ("Widen caps"), so the name and shape follow existing precedent.
- The handler reads it once and passes it through in place of the six literals. Defaults do not change, so default token cost does not change.
- Tool-level tests mirroring `full_flag_raises_match_cap`: default at most 10, full more than 10, for symbol, content and callers.

Opening this first per CONTRIBUTING since it touches a tool schema. I will follow with a PR of that shape; if you would rather have a numeric `limit`, say so and I will adjust.

jahala:
Reviewed by Fable 5.1

Boolean was the right shape; it mirrors `tilth_grok` and costs the schema one property. Landed in #217.

PR #217, opened 2026-09-12
feat(mcp): let tilth_search raise its own result cap

plusky:
Every search path behind `tilth_search` already honours a `full` flag that raises the match cap from 10 to 100 — `content.rs`, `symbol.rs` and `callers.rs` each pair `FULL_MAX_MATCHES = 100` with a proportional walker early-quit threshold, and `--full` reaches them from the CLI. The MCP handler did not: it passed the literal `false` in all six dispatches and the schema had no such property, so ten results was a hard ceiling for an agent no matter which kind it asked for. This implements #215.

The change reads `full` once in `tool_search`, next to `expand`, threads it into the six dispatches, and adds the schema property in the shape `tilth_grok` already uses. The cap is per query, not per call: both multi arms apply it inside their own loop (`search/mod.rs:214`, `search/callers.rs:592`), so a five-target comma search under `full` can render up to 500 entries. The schema wording says "per query" so the number is not read as a total.

Five tests in `src/mcp/tools/search.rs`, all at the tool boundary. Coverage below the tool layer is thinner than it looks — `symbol.rs`'s `full_flag_raises_match_cap` is the only full-cap test in the search layer, and it covers the symbol path alone; content and callers have none. Each test builds a tempdir of 15 files with one match each and counts how many of them the output names, rather than matching a header phrase, so the assertions survive header and facet rewording. Defaults are asserted as exactly 10, not at most 10, which is deterministic here because 15 hits sit under every early-quit threshold on these paths. The symbol and callers tests each exercise their single and comma dispatches; the regex test queries `Needle[A-Z][0-9]` against files containing `NeedleQ7`, so it matches only under regex semantics and pins the `is_regex` argument sitting beside `full` at that call.

Mutation-checked: restoring each of the six literals in turn fails exactly one of the new tests and nothing else, and swapping the two adjacent bools on the regex line — which type-checks and would silently turn regex search into literal search at the raised cap — fails the regex test with `— 0 matches`.

Measured, on a 120-hit tree: at the default `expand=2` a `full` answer renders all 100 matches in roughly 1.8k tokens for short definitions and 6.8k for 200-character content lines, comfortably inside `DEFAULT_BUDGET` (24k). `budget` only bites when the raised cap meets wide expansion; at `expand=100` over large function bodies the output hits a ceiling around 95 kB / 23.8k tokens and the allocator drops the lowest-value entries, saying so on its own line (`... N lower-value match(es) omitted to fit budget`).

The cost is on the schema, not the responses. The new property adds 188 bytes to every `tools/list` — 8669 to 8857 read-only, 12327 to 12515 in edit mode, about 47 tokens by `estimate_tokens` — and every session pays that. A call that omits `full` returns byte-identical output to today, since the default reproduces the literal it replaced. Worth an A/B if the prompt-size budget is tight.

One adjacent issue, out of scope but relevant to anyone raising this cap. Two behaviours that are individually deliberate and commented — `content::search` leaving `FacetTotals` all-zero (`types.rs:146-148`) and the faceted render path suppressing the global "and N more" tail to avoid double-counting (`search/mod.rs:1191-1193`) — combine so that a capped `kind=content` answer discloses nothing: a 15-hit search prints `— 10 matches` under a bare `## Usages — other (10)` and no tail. `kind=callers` does print a pre-truncation total in its header, but that total is itself floored by the walker's own early-quit budget. Raising the cap widens the window where that silence matters; fixing it belongs in its own change.


jahala:
Reviewed by Fable 5.1

I read the full diff and ran fmt, clippy `-D warnings` and the test suite on stable 1.98.1, all green. Verified over a real MCP round trip on tilth's own tree: `kind=content MAX_MATCHES` answers `10 of 24` by default and all 24 with `full`. The 47-token schema cost is acceptable for a property that removes a grep fallback. #219, #220 and #218 stack on this cleanly. Merging with Jan's approval.

PR #218, opened 2026-09-12
feat(lang): see Rust calls inside macro and attribute arguments

plusky:
**What changed**

tree-sitter parses a macro invocation's arguments as a `token_tree` of raw tokens (`grammar.js:225`, `:978`), so a call written inside `write!(..)`, `vec![..]` or `assert!(..)` has no `call_expression` node and Rust's callee query never saw it. Over tilth's own `src/`, `format_token_count` reported 1 call site where there are 3, and `estimate_tokens` 32 where there are 43.

The new pattern is `(token_tree (identifier) @callee . (token_tree . "("))`. Requiring the following tree to open with `(` is what separates a call from `x[i]` and `Foo { .. }` — without it the capture set over `src/` grows from 2297 to 2488 and picks up `else` from `if (c) {1} else {2}`. But the anchor between the identifier and the tree skips anonymous nodes, so the pattern still matches the `x` in `assert_eq!(x, (a, b))`, where only a comma stands between. Tree-sitter has no way to say "and no anonymous nodes either", so the rest is a predicate: `LangSpec` gains `callee_filter: Option<fn(&tree_sitter::Node) -> bool>`, beside `extract_receiver` and in the same optional-fn-pointer style, applied by the two `@callee` consumers (`callers::find_callers_treesitter_batch`, `callees::extract_callee_names`). Rust's accepts a capture whose parent is a token tree only when the argument tree is its immediate next sibling; captures from the other four patterns hang off `call_expression` / `field_expression` / `scoped_identifier` / `macro_invocation` and pass untouched. The other 18 specs set `None`.

Because the pattern keys on token trees rather than on macros, it reaches every token tree: macro arguments, `macro_rules!` expansion bodies, and attribute argument lists. That last one is deliberate — `#[error("not found: {}", path.display())]` is a thiserror format argument that really is compiled into the `Display` impl, and on tilth's own source 9 of the 15 attribute-tree captures are calls of exactly that kind. The price is that `cfg` predicates have the same shape.

Four tools change, not one: `tilth_search kind=callers`, the callee direction (`grok`, `tilth_deps` via `deps.rs:118`) and `tilth_diff`'s blast radius (`blast.rs:85`). Measured: `tilth src/mcp/tree.rs --deps` gains three `→ fmt_tokens` rows (0 → 3).

One adjacent change in `callers.rs`: `call_text` rendered a call whose parent spans lines as the bare symbol name — a useless echo of the query, and a macro's token tree nearly always spans lines. It now falls back to the callee's own source line. **This is not Rust-specific**: every language's wrapped call sites now render that line instead of the name (Python `-> target` becomes `-> return target(`). Scoping it to token trees would mean hard-coding a Rust node kind inside language-agnostic code, which is the coupling `LangSpec` exists to remove.

**How to test**

`cargo test`. Six new tests: `rust_callee_query_finds_calls_in_macro_arguments` (compiles the query explicitly — `with_callee_query` swallows a compile error with `.ok()?`, so a malformed pattern would silently disable callers and callees for Rust with no error anywhere), `extract_rust_callee_names_inside_macro_arguments` (positive, plus negatives for the filter and for the `(` requirement), `extract_rust_callee_names_inside_attribute_arguments`, `callers_sees_calls_inside_macro_arguments`, `call_text_falls_back_to_the_callee_line_across_languages`, and `only_rust_filters_callee_captures`.

Each was checked against a hand mutant: removing the pattern, removing the filter, dropping the `(` requirement, reverting the `call_text` hunk, and excluding attribute trees each kill the corresponding test. `only_rust_filters_callee_captures` cannot compile without the field, so it was proven by the filter-removed run instead.

**Known limits**

The pattern matches an identifier-shaped token immediately followed by a parenthesised tree, and raw tokens are all there is to go on:

- Patterns read as calls: `matches!(x, Foo(_))` yields `Foo`, indistinguishably from the nested call in `vec![Some(Wrapped("y"))]`.
- Keywords read as calls: `m!(for x in (0..2) {})` yields `in`, `m!(move (1))` yields `move`, `m!(&'a (x))` yields the lifetime `a`, `asm!("nop", options(nostack))` yields `options`. None occur in tilth's own source.
- `cfg` predicates read as calls: `#[cfg(not(any(unix, windows)))]` yields `not` and `any`, attributed to `<top-level>`.
- A call in a `macro_rules!` expansion body is a call site attributed to the enclosing item, even for a macro that is never invoked; the code that invokes the macro is still not reported. (The LHS `token_tree_pattern` is correctly skipped.)
- A turbofish inside a macro is missed: `m!(foo::<T>(x))` captures nothing, because `foo`'s next named sibling is `T`. Consistent with pre-existing behaviour.
- The callee direction drops names with no definition in scope, so most of the above never reach `grok` or `deps` output.

**Cost**

Real trees are unaffected: tilth's own `src/` is 0.27 s before and 0.28 s after. The new pattern is quadratic in bracket-nesting depth inside a token tree — synthetic depths of 500/1000/2000/4000 cost 0.16/0.25/0.77/2.66 s against a flat 0.10–0.16 s baseline. Reaching that needs hostile or generated input; across 52,960 real registry `.rs` files the deepest nesting is 185, where the delta is noise.


jahala:
Reviewed by Fable 5.1

Full diff read; fmt, clippy `-D warnings` and tests green on stable 1.98.1. Confirmed on tilth's own tree: `format_token_count --callers` goes from 1 to 3 sites, and the cfg predicates show up as described (`not` reports 4 sites, all `#[cfg(not(..))]`). That noise is a fair price for what the pattern buys `grok` and `deps`. Merges cleanly on top of #217, #219 and #220.

One non-blocking note: `only_rust_filters_callee_captures` pins today's fact rather than a contract, so the next language that needs a filter has to edit it. Keep or drop, your call. #216 covers the same macro case with a looser pattern and will rebase onto this. Merging with Jan's approval.

PR #219, opened 2026-09-12
fix(cli): keep --full's raised match cap when --expand=0

plusky:
`tilth <query> --full --expand=0` returned fewer matches than `tilth <query> --full`, and exactly as many as passing no flag at all. On this repo: `tilth MAX_MATCHES --full --glob '*.rs'` reported 15 matches, adding `--expand=0` dropped it to 10.

`--expand=0` is meant to cancel inline source and nothing else, but it also sends main down `lib::run_full`, which passed `cli_full: false` into `run_inner`. `run_inner` then reached `run_query_basic`, which took no `full` parameter at all, so the six `search` wrappers it calls each supplied their own hard-coded `false`. The 10 -> 100 cap raise was therefore reachable only through the expanded path. The output did disclose the truncation ("## Usages — same package (7/12)" and "... and 5 more usages"); what it never said was that `--full` had been dropped on the floor.

`run_full` now takes the parsed `--full` flag, `run_inner` forwards it to `run_query_basic`, and that forwards it to all six entry points the unexpanded path can reach — `search_symbol`, `search_content`, `search_regex`, and the `_raw` trio behind the single-word-concept, fallthrough and multi-word cascades. All six were capped at 10 before, so threading only the obvious three would have left `tilth widget --full --expand=0` and `tilth "tag 424242" --full --expand=0` broken.

A seventh path changes without appearing in the diff: the multi-symbol branch (`lib.rs:265`) already forwarded `cli_full`, so its value flips too — `tilth "MAX_MATCHES,FULL_MAX_MATCHES" --full --expand=0` goes from 10/10 matches to 15/12. That branch also rewrites `expand` 0 -> 2, so it is the one shape where `--expand=0` still emits inline source; the help text says so.

`run_full` is also how a piped invocation arrives (`full = !is_tty` with no `--full` typed), so it receives `cli.full`, never the derived bool. The guarantee pinned by `piped_invocation_does_not_auto_expand` is unchanged and now has a companion test one layer down, plus a comment at the call site, since a one-token slip there leaves the suite green.

This completes cc64a42 (#111), which raised the cap for the expanded path and for `--callers` but left every unexpanded dispatch behind. Like that commit, this adds a required parameter to a published `pub fn` (`run_full`), which is source-breaking for a crate depending on `tilth` as a library; cc64a42 made the same change to `run_expanded` and shipped in the v0.8.3 patch release. In-repo the only caller is `main.rs`; `search` is `pub(crate)`, so the six wrapper signatures are invisible outside.

`search_symbol_raw` is also used by grok, which passes `false` explicitly. That is not a no-op choice: `full` raises the definitions early-quit threshold from 50 to 300, and on a 62-definition fixture the plain search collected 53-55 definitions across 10 runs against a steady 62 with the flag. Keeping `false` leaves grok byte-identical.

How to test:

```
cargo test                      # 652 lib tests, +3
cargo build
./target/debug/tilth MAX_MATCHES --full --glob '*.rs' | head -1
./target/debug/tilth MAX_MATCHES --full --expand=0 --glob '*.rs' | head -1
./target/debug/tilth MAX_MATCHES --glob '*.rs' | head -1
```

Both `--full` forms now report 15; the un-flagged piped run still reports 10.

Measured. Over the nine query shapes that reach the unexpanded dispatch, on a 15-file fixture (with `--full` / without): every shape returned 10/10 before; after, 15, 30, 15, 30, 30, 15, 15, 15 and 15+12 against a steady 10. On a 120-definition fixture, `--full --expand=0` lists 100 matches with zero fenced blocks while bare `--full` lists the same 100 with 50 blocks. Wall clock on a wide query over tilth's own tree, best of three: 339 -> 816 ms, still under bare `--full` at 967 ms; on a narrow query 502 -> 540 ms.

One behaviour worth flagging: with a small `--budget`, the larger result is now trimmed by `budget::apply`'s positional byte-cut rather than returned whole, because the unexpanded path passes no budget into `fit_to_budget`. At `--budget 800` a wide query used to return a complete 10-item list and now returns a truncated slice of 100. That is the pre-existing shape of the unexpanded path, not something this change introduces, but the raised cap makes it reachable.

Documentation: neither `--full` nor `--expand` mentioned the match cap, and both described the expansion as "every match" when it stops at `FULL_EXPAND_CAP = 50`. `skills/SKILL.md` repeated the same claim. All three surfaces now state the 10 -> 100 cap, the 50-match expansion limit, that `--expand=N` sets the inline-source count only, and the multi-symbol exception.


jahala:
Reviewed by Fable 5.1

Full diff read; fmt, clippy `-D warnings` and tests green on stable 1.98.1. Reproduced before and verified after on tilth's own tree: `MAX_MATCHES --full --glob '*.rs' --scope src` reports 14, `--full --expand=0` now also 14, bare stays at 10. The `run_full` signature change follows the precedent `run_expanded` set in 0.8.3. The doc corrections on the 50-match expansion cap were overdue. Merging with Jan's approval.

PR #220, opened 2026-09-12
fix(search): disclose the display cap, bound what content collects

plusky:
This ports the disclosure and walk-completion halves of your own `d67fc1b`, narrowed to the content/regex path so it does not wait behind the scout work in #169. The design and the wording in `search_header` — including `(the #51 bug)` — are yours. One part is not yours, and it is the part I most want you to look at.

**Please read this before you land either version: `d67fc1b`'s content.rs hunk, and the identical hunk in #169, remove the only thing bounding how much a content or regex search collects, and the shape of the cost is not the walk.** A complete walk over 278 MB with no hits is 17 ms. What the quit was bounding is the `Vec<Match>` — one `PathBuf` plus one `String` per matching line — and `rank::sort`, whose comparator recomputes `score()` O(n log n) times. Measured through the real MCP server on a 278 MB / 14,743-file registry tree, `kind=content "e"`: base 0.46 s / 12 MB, your patch 379 s / 1.0 GB. `kind=regex "."` does not answer inside the 90 s default; `timeout.rs` detaches the worker rather than cancelling it, so it keeps running and keeps allocating, and after `MAX_ABANDONED_THREADS` = 8 the server refuses every subsequent tool call. Eight such calls peaked at 11.5 GB and left the server answering "server busy". Scope and pattern both come from the agent, and `kind=regex` makes a high-frequency pattern a one-line request.

So this PR adds a replacement bound. Each file now contributes at most `max_matches` entries, chosen by the same score and the same (score, path, line) tie-break `rank::sort` uses; `total_found` is counted separately and stays exact. It is lossless, not an approximation: the comparator restricted to one file is the same total order, so anything in the global top-N has at most N-1 predecessors in its own file and is therefore in that file's top-N. The bound is per file rather than global on purpose — a global cap would put the thread-timing-dependent result set back, which is the thing this change exists to remove. `rank::retain_top_k` also scores each match once instead of O(n log n) times, which is where most of the time went.

I deliberately did not touch `symbol.rs` or `callers.rs`, or the `diff.rs` / `deps.rs` / `blast.rs` hunks of `d67fc1b`.

**What changed**

- `search_header` takes `shown` as well as `total` and renders `"{shown} of {total} matches"` when the cap bit. This changes capped symbol headers too; `callers` has its own header and is untouched.
- On the faceted path, a result whose per-facet totals are all zero emits the global `... and N more matches` tail. That is content and regex; symbol results always have a non-zero total because `facet_matches` buckets every match, so the two tails can never both appear.
- The content walker runs to completion, and collection is bounded per file.

**How to test**

`cargo test` — six new tests. The bound is pinned by `retain_top_k_keeps_exactly_what_a_full_sort_keeps` (a fixture where the highest-ranked line is the *last* one, so a "first k per file" bound would drop it) and by `per_file_cap_bounds_collection_without_changing_results`, which runs the same fixture at the production bound, at a raised bound and at one-per-file: the exact total and the displayed set are identical across the first two, and the third shows the bound is live. The disclosure half is pinned end-to-end on the content path and, for the symbol path, by a test that a capped symbol result carries its per-facet tail and never the global one, and that each multi-symbol section header discloses its own cap. Every one of them was proven to fail against the corresponding mutant.

Manually: `tilth 'let mut' --scope src/` heads with `10 of 510 matches` and ends with `... and 500 more matches`.

**What was measured** (MCP `tools/call`, release binaries, wall / peak RSS — base → `d67fc1b` as-is → this PR)

| corpus | query | base | d67fc1b | this PR |
|---|---|---|---|---|
| registry, 278 MB | `content "e"` (2.87 M hits) | 0.46 s / 12 MB | 379 s / 1019 MB | 18.7 s / 67 MB |
| " | `regex "."` (5.50 M hits) | 0.32 s / 12 MB | 90 s timeout, 1672 MB climbing | 22.3 s / 87 MB |
| " | `content "error"` (24 k hits) | 0.33 s / 12 MB | 2.63 s / 18 MB | 1.55 s / 15 MB |
| windows-0.62.2, 111 MB | `regex "[a-z]"` (712 k hits) | 0.09 s / 12 MB | 115 s / 306 MB | 1.66 s / 34 MB |
| tilth `src/` | `content "e"` (23.7 k hits) | 0.08 s / 12 MB | 1.50 s / 14 MB | 0.13 s / 12 MB |

Eight heavy `regex "."` calls followed by a cheap one, one session, default timeout: with `d67fc1b` every heavy call times out and the cheap one is refused; with this PR all eight answer (175 s total, 204 MB peak) and the cheap one succeeds. Determinism: ten identical calls give 2 and 4 distinct outputs on base, 1 and 1 now. Bounded output is byte-identical to the unbounded build on `content "e"` and `regex "."` over tilth `src/` and `regex "let"` over the 111 MB crate.

**Things I want to flag rather than hide**

- The bound raises the bar for the abandoned-thread wedge but does not remove it. Force `TILTH_TIMEOUT=1` and eight slow calls still reach 497 MB and still refuse the ninth. That is `timeout.rs` detaching workers, not the walker; it may deserve its own issue.
- Capped symbol results also change their `(~N tokens)` footer by a couple of tokens (`~497` → `~499`), which is what `fit_to_budget` works against, so a result sitting exactly at `DEFAULT_BUDGET` could lose a segment. I did not add a boundary test for that; pinning the estimator seemed worse than the risk.
- Capped content/regex token cost moves in both directions, because the candidate pool is now complete and different matches rank into the top ten (`content "let"` 987 → 813; `regex "Ordering"` 337 → 435). Only uncapped output is byte-identical.
- Two pre-existing cosmetics are now routine on the content path rather than rare: `... and 1 more matches.` reads wrong for a single hidden match, and the facet heading still shows a bare `## Usages — other (10)` for 15 hits because `count_label` sees the zero total. Both predate this change and I left the wording alone.
- The honest header also makes the symbol walker's remaining raciness visible in the first line — six identical `tilth walker` runs print `10 of 38`, `36`, `36`, `46`, `32`, `33`. It was already visible in the `9/31` facet label, but it is a good argument for landing the rest of your commit, with a bound on those walkers too.


jahala:
Reviewed by Fable 5.1

Full diff read; fmt, clippy `-D warnings` and tests green on stable 1.98.1. The per-file top-k argument is sound (same total order, so the global top-N is a subset of the per-file top-Ns), and the tests pin the case a first-k bound would fail. Verified on tilth's own tree: `'let mut' --scope src` heads `10 of 528 matches`, tails `... and 518 more`, identical across runs. The large-tree measurements are yours and I did not reproduce them, but the direction is clear: Jan's d67fc1b content hunk in #169 will be dropped in favour of this.

Two follow-ups noted for the maintainer: the symbol and callers walkers need the same bound, and `rank::sort` can score once the way `retain_top_k` does. The `timeout.rs` abandoned-thread wedge deserves its own issue. You have the numbers, please open it. Merging with Jan's approval.

